PR #87, opened 2023-01-06
[Tech] Add Support for Lottie Animation Files

BrettCleary:
This adds lottie animation support to the frontend. Currently in draft as some Lottie animations (from https://lordicon.com/) give the following error
```
this.elements[i].destroy is not a function
    at SVGRendererBase.destroy
```
which originates from the `"lottie-web"` package

The working test animation is from https://lottiefiles.com/132124-hands-typing-on-keyboard.

This should remain in draft until designers need to use lottie animation files

---


BrettCleary:
This logic should be moved to the ui library

PR #157, opened 2023-02-24
[Tech] Enable Auto Updates

BrettCleary:
Changes urls for auto updater and adds release workflows for all platforms

Now on new tag push, github will automatically build for all platforms and draft a release

### Screenshots
image
image

---


.github/workflows/release_linux.yml
```diff
@@ -0,0 +1,41 @@
...
+on:
+  push:
+    tags:
+      - 'v*'
+  workflow_dispatch:
+
+env:
+  GITHUB_TOKEN: ${{ secrets.WORKFLOW_TOKEN }}
+  GH_TOKEN: ${{ secrets.WORKFLOW_TOKEN }}
+  CSC_IDENTITY_AUTO_DISCOVERY: true
```
flavioislima, line 12: Those should not be needed, they are only for macOS signing.
We just need the github ones.

.github/workflows/release_linux.yml
```diff
@@ -0,0 +1,41 @@
...
+jobs:
+  draft-releases:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository.
+        uses: actions/checkout@v3
+        with:
+          submodules: 'recursive'
+          token: ${{ secrets.pat }}
+      - name: Setup NodeJS
```
flavioislima, line 28: I think we can use v3 of all workflow and update node to 18.

.github/workflows/release_windows.yml
```diff
@@ -0,0 +1,41 @@
...
+  push:
+    tags:
+      - 'v*'
+  workflow_dispatch:
+
+env:
+  GITHUB_TOKEN: ${{ secrets.WORKFLOW_TOKEN }}
+  GH_TOKEN: ${{ secrets.WORKFLOW_TOKEN }}
+  CSC_IDENTITY_AUTO_DISCOVERY: true
+  CSC_LINK: ${{ secrets.CSC_LINK }}
```
flavioislima, line 13: Same here, not needed

package.json
```diff
@@ -201,7 +201,7 @@
     "manualTestProxyJest": "jest mmExtIntegration.test.ts",
     "release:linux": "vite build && electron-builder -p always --linux deb AppImage rpm pacman tar.xz",
```
flavioislima, line 202: I think we should also remove AppImage (since it has a bug with the mm extension that loses configuration on restart)

src/backend/main.ts
```diff
@@ -256,7 +256,11 @@ const loadMainWindowURL = function () {
       `file://${path.join(publicDir, '../build/index.html?view=App')}`
     )
     if (!isMac) {
```
flavioislima, line 258: We should remove this check. On heroic it was giving an error message since we do not have a certificate for signing, but here should be fine

flavioislima (approved):
LGTM

BrettCleary:
I re-enabled the husky prepush command. In the future we can circumvent this with `git push --no-verify` https://github.com/typicode/husky/issues/124

PR #163, opened 2023-02-27
[FIX] Sidebar Drag fix + Pull fixes from Heroic

flavioislima:
This fixes the sidebar dragging not working on webviews and some other fixes and improvements from Heroic.

---

### DIFF
--- a/src/frontend/screens/WebView/index.tsx
+++ b/src/frontend/screens/WebView/index.tsx
@@ -15,10 +15,6 @@ import { Runner, WebviewType } from 'common/types'
 import './index.css'
 import LoginWarning from '../Login/components/LoginWarning'
 
-type CODE = {
-  authorizationCode: string
-}
-
 export default function WebView() {
   const { i18n } = useTranslation()
   const { pathname, search } = useLocation()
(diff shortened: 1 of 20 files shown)


src/frontend/screens/WebView/index.tsx
```diff
@@ -193,6 +204,7 @@ export default function WebView() {
         src={startUrl}
         allowpopups={trueAsStr}
         useragent="Mozilla/5.0 (Windows NT 10.0; WOW64; rv:70.0) Gecko/20100101 Firefox/70.0"
+        {...(preloadPath ? { preload: preloadPath } : {})}
```
BrettCleary, line 207: We should call setPreloads on a persistent session and then load the webview into that session in the future. It would create a better separation between frontend and backend,  also eliminate a lot of boilerplate, and be more easily extensible.

BrettCleary (approved):
The sidebar resizing is working for me. I did not time to test much else, but noticed no issues while testing

src/frontend/screens/WebView/index.tsx
BrettCleary: we can refactor this later if we end up needing another webview with the same preload

PR #187, opened 2023-03-13
[Tech] MetaMask update & migration

hellowodl:
feat: added logic to migrate older extension versions to newer ones

feat: added MetaMask v10.25.0

In order to test this, it'd be prudent to backup your ~/.config/hyperplay directory so that test can be restarted.
Subsequent updates can be triggered by copying & renaming the /public/extension/MetaMask_v10.25.0 directory to a higher version.

---


BrettCleary:
the extension id is an SHA hash of the absolute path to the extension 
https://stackoverflow.com/questions/26053434/how-is-the-chrome-extension-id-of-an-unpacked-extension-generated

so if we delete the old extension and name the v10.25.0 extension `MetaMask_v10.22.2`, then it will generate the same hash and have its level db in the same place locally. No migration needed.

I think we should do this instead of introducing this migration logic

BrettCleary:
> the extension id is an SHA hash of the absolute path to the extension https://stackoverflow.com/questions/26053434/how-is-the-chrome-extension-id-of-an-unpacked-extension-generated
> 
> so if we delete the old extension and name the v10.25.0 extension `MetaMask_v10.22.2`, then it will generate the same hash and have its level db in the same place locally. No migration needed.
> 
> I think we should do this instead of introducing this migration logic

we should update the ReadMe with this info

BrettCleary:
Okay I patched some mm issues and updated the readme

Tested:
- connect wallet
- switch accounts
- window.ethereum.on methods
- /callcontract /sendcontract /rpc methods

hellowodl:
I've tested:
 - Ingame transactions(thusfar only a network change within defikingdoms)
 - importing a wallet from chrome
 - importing a wallet with a seed phrase
 - migrating a wallet from 10.22.2
 - creating a new wallet

All seems to work flawlessly.

As I originally created the PR, I can't approve it, though it certainly has my blessings.

@BrettCleary

BrettCleary (approved):

PR #188, opened 2023-03-13
[Tech] Added MM darkmode & fixed browser import issues

hellowodl:
* Fixes an issue in which the user can't import MetaMask wallets from other browsers
* Adds dark mode as the default setting for newly created/imported MetaMask wallets 

---


BrettCleary:
For `Fixes an issue in which the user can't import MetaMask wallets from other browsers` can you add steps to reproduce?

BrettCleary (approved):
LGTM

PR #202, opened 2023-03-14
[Fix] Do not show Anti-cheat info for HP games

flavioislima:
Skip check for HP games.

---


BrettCleary:
The issue here is that the backend erroneously returns fortnite for all HP games due to the namespace check being `.includes` and HP game namespaces being empty. We should fix this on the backend instead of patching like this every time it is called on the frontend. HP games might be added to are we anticheat yet sometime in the future as well.

I will close this branch as I have a fix here: https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/pull/203 that also has a good performance improvement as well

PR #208, opened 2023-03-17
feat: overhauled & animated extension popup and notification

hellowodl:
* feat: clicking outside of popup closes the popup(if the clicked element isn't a webview)

* fix: 2 popups showing up at the same time

* fix: popup menuitem temporarily doesn't react to clicks

* fix: popup menuitem at times doesn't show whether the popup is open or not

A small store has been introduced that manages the popup open state, and arranges the communication between the sidebar & ExtensionManager. Having a single source of truth reduced complications and clarifies the logic as opposed to having this communication run through the backend.

---

### DIFF
--- a/src/frontend/ExtensionManager/index.module.scss
+++ b/src/frontend/ExtensionManager/index.module.scss
@@ -1,31 +1,30 @@
 .mmContainer {
+  // @include proportions;
+
   position: absolute;
-  left: 250px;
-  top: 10px;
+  right: 550px;
   display: flex;
   flex-direction: row;
   gap: 20px;
   z-index: 100000;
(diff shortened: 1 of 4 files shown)


flavioislima (approved):
Tested here and looks good.
I think one small improvement would be to handle the mouse click when on a webview, right now the popup does not close if you click outside when on a webview like the portfolio or any store.

hellowodl:
@flavioislima I've indeed noticed the same issue. I've investigated a workaround though it appears that the click event has to be caught from the context of the webview, not outside of it. Because of that another issue has been opened for this, as it's a longer & more complex task.

PR #212, opened 2023-03-17
[Tech] Refactor Protocol and add support for HP Games

flavioislima:
This PR refactors the protocol handler so it is easier to extend and also changes the default way we create shortcuts so we add a runner on the command instead of guessing the runner by looping through all of them to figure it out, so it is more optimized. But it will continue working for current shortcuts.

This also adds support for HP games, it is possible to install and launch games.
If a game is not on the library, it also adds automatically to it.
It also changes how we ask the user to install, instead of showing just the folder selector, now it will show the Install dialog just like a normal install, except for Browser games that will be launched right away.

With that, any developer can add a button to their website called `Play on Hyperplay` that will call: `hyperplay://launch/hyperplay/game_id`, for instance, and then it will open HP, add the game to the library and launch or install first.

### How to test Installed games
- install one of the builds or build your own installer (won't work on dev mode)
- Go to any installed game and click on the three dots and Add Shortcuts
- Launch from Shortcut. The game should launch just fine.

### How to test add games to the library from HP store
- Check on the API for a game ID that is not on your current library.
- Run HP from the terminal with the following command: `/path/to/binary hyperplay://launch/hyperplay/game_id`
- Hp should open and add the game to the library and show the install modal in case it is not a browser game. If it is a browser game it should just launch it.

### Caveats
- Browser games launched like this think that should connect using the MM mobile so I could not find a way of connecting it to the MM extension without opening HP and doing so.

---

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -514,8 +514,11 @@ ipcMain.once('frontendReady', () => {
 
 // Maybe this can help with white screens
 process.on('uncaughtException', async (err) => {
-  logError(`${err.name}: ${err.message}`, LogPrefix.Backend)
-  showDialogBoxModalAuto({
+  logError(
+    `'Uncaught Exception occured!': ${err.name}: ${err.message}`,
+    LogPrefix.Backend
+  )
+  /*   showDialogBoxModalAuto({
     title: i18next.t(
       'box.error.uncaught-exception.title',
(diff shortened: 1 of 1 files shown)


hellowodl:
Seems to work great.
LGTM

hellowodl (approved):

PR #214, opened 2023-03-18
[Fix] import mm crash

hellowodl:
Fixes an issue of the application crashing when trying to install the MetaMask extension.

---


BrettCleary (approved):
good fix for when there are no browsers to import from

PR #217, opened 2023-03-18
[Fixes] Final 0010 fixes

BrettCleary:
Some final fixes before releasing v0.0.10

This fixes:
- importGame for HP games was causing overlay injection to fail due to an incorrect install platform value
- importGame was not setting version, causing all HP imported games to show that they needed an update
- on app launch, HP games that are out of date would not show the update button until refresh was clicked
- on app launch, a white background color would be shown on the update component screen and also when resizing the app

Also
- updates discord link to discord.gg/hyperplay

---


BrettCleary:
tests are failing once again due to `ipcMain.once` in `storeManagers/index.ts`. This is because we aren't mocking the electron calls correctly and should be revisited on a separate branch

hellowodl (approved):

PR #230, opened 2023-03-27
[FIX] Platform Handlers on Launch and on GamePage

flavioislima:
- Fix native games being misidentified as Windows games on macOS.
- Fix Browser games and native macOS games showing crossover/wine information on game page.
- Fix HP games not showing on favorites.
- Set `is_mac_native` and `is_linux_native` when adding games to the library. (should be fixed for current users once we have the refresh method as well)

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hyperplay",
-  "version": "0.0.9",
+  "version": "0.0.10",
   "private": true,
   "main": "build/electron/main.js",
   "homepage": "./",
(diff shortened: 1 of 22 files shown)


BrettCleary:
Code looks good. Thanks for fixing this.

I just tested on Windows
- Adding a hp game to favorites
- Adding a browser hp game and refreshing to see if update icon shows

All working for me. The logic on Mac/Linux seems simple enough and noncritical so I will go ahead and approve.

I think at some point we need to refactor the `InstallPlatform` type in `src/common/types/ts`. That type was covering the frontend types for platform and also the hp store types through union and that's why we didn't catch this issue initially in `src/backend/storeManagers/hyperplay/games.ts` 

We can do this on a different PR

BrettCleary (approved):

PR #233, opened 2023-03-28
[Fix] Import Folder

BrettCleary:
Import directory is necessary for single file applications on mac so this reverts the previous change to import executable

---

### DIFF
--- a/public/locales/en/gamepage.json
+++ b/public/locales/en/gamepage.json
@@ -6,7 +6,7 @@
             "title": "Change Games Install Path"
         },
         "choose": "Choose",
-        "importpath": "Choose game executable to import",
+        "importpath": "Choose Game Folder to import",
         "move": {
             "message": "This can take a long time, are you sure?",
             "path": "Choose where you want to move",
--- a/src/backend/storeManagers/gog/games.ts
+++ b/src/backend/storeManagers/gog/games.ts
@@ -12,7 +12,7 @@ import {
   getGameInfo as getGogLibraryGameInfo,
   changeGameInstallPath
 } from './library'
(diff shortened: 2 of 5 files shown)


flavioislima (approved):
LGTM

BrettCleary:
> LGTM

I just tested importing carto demo (gog game) which is a single app file on mac and I couldn't select it. Hundred days on epic with mac also has the same issue.

But if we change it to import the executable, I believe legendary/gog will fail since currently we pass the folder path as an import param for games that are shipped in a folder.

How can we get both of these cases working? @flavioislima

BrettCleary:
> > LGTM
> 
> I just tested importing carto demo (gog game) which is a single app file on mac and I couldn't select it. Hundred days on epic with mac also has the same issue.
> 
> But if we change it to import the executable, I believe legendary/gog will fail since currently we pass the folder path as an import param for games that are shipped in a folder.
> 
> How can we get both of these cases working? @flavioislima

This is a known issue on Heroic and seems to affect a small % of games. I will leave it for now

flavioislima:
> > > LGTM
> > 
> > I just tested importing carto demo (gog game) which is a single app file on mac and I couldn't select it. Hundred days on epic with mac also has the same issue.
> > 
> > But if we change it to import the executable, I believe legendary/gog will fail since currently we pass the folder path as an import param for games that are shipped in a folder.
> > 
> > How can we get both of these cases working? @flavioislima
> 
> This is a known issue on Heroic and seems to affect a small % of games. I will leave it for now

Yes, I just fixed it on heroic so I can cherry pick it here later 👍

PR #249, opened 2023-04-05
[Feat] Show Overlay by Default

BrettCleary:
Show overlay automatically on notification so the user does not have to toggle on the overlay for each event

Refactors `overlay.ts` to use a module level interface instead of a singleton within a module which is hard to reason with, test, and was giving errors when toggling overlay state.

---


hellowodl (approved):
This seems to work well, though I've thusfar only tested a browser game, not a native game.

PR #287, opened 2023-04-27
[Feat] Pause Downloads Queue

flavioislima:
Pulled the PR from heroic that introduced the pausing for the Downloads Queue.
Refactored the whole download part for the HP games to use [EasyDL](https://github.com/andresusanto/easydl) that is a lot more complete and have more features and is also more reliable than Axios in general for that matter.

> TODO: Refactor all other places that we use Axios or other tools to download (Wine/proton/winetricks/dxvk) to use this lib as well.

- [x] Pull the heroic feature of pausing the downloads queue for Epic and GOG.
- [x] Implement Pause downloads for HP Games.
- [x] Add Pause/Continue button on the game card
- [x] Add Continue button on game page

---

- [x] Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)


flavioislima:
@stormcakes @BrettCleary 

How to test:

- Start a download
- go to the Downloads screen and click on the pause button. The whole queue should be paused (not only the first download)
- Click play and the download should continue.
- Try this with games from Epic, GOG and HyperPlay.
- Start downloading a HyperPlay game. 
- Check for temporary files under: `${configFolder}/.temp`
- Click on Stop and keep downloaded files.
- Then click install again, the game should start downloading from where it stopped.
- Try again but this time select NOT to keep the downloaded files
- The temporary files should have been deleted.

The temporary files for EPIC are on the selected install folder.
The temporary files for GOG are on the cache folder on the config folder.

Config Folder:
Linux: `~/.config/hyperplay`
Windows: `%AppData%\hyperplay`
macOS: `~/Library/Application Support/hyperplay`

Let me know if you have any other further questions 👍🏽

BrettCleary:
Okay I tested and had some issues but it was very cool seeing the hp download pause and resume. Nice job finding this package and integrating it @flavioislima


- I noticed that the card shows not available and a play button if paused. We probably need a paused state added
- I get a windows installed notification when paused is clicked
- the .temp folder still exists after the installation finishes with the full 1.5 GB .zip file (tested with another world). I see that you have some rmSync calls but it doesn't seem like they are getting called. either that or the folder is wrong


1. I start download then click stop and keep installed files while downloading moonblasters but the game card shows as installed with a play button and everything
2. clicking play shows an error that exe cannot be found
3. I do see the moonblaster files in .temp though
4. when I click uninstall the files are still there even when I click remove log and settings (I suppose this is expected behavior though)
5. when I click install, it starts from where it left off (nice!)
6. stopping and keeping files gives the same issue above with the game card showing as installed but the game gives a "cannot find exe" error when clicking play because it isn't actually downloaded
7. I resumed and stopped the download, clicking do not keep files and the files were still there in .temp and moonblasters showed as installed like in 5 where clicking play errors again because it isn't actually downloaded

flavioislima:
Thanks for testing @BrettCleary , I did not get some of these issues while developing but I will test more to try to reproduce it and will send a fix soon.

flavioislima:
Yeah, on macOS if I keep the files the games still show as not installed, both windows and macOS games.
I also get the correct notification when pausing:
image

Will open windows to test now.

flavioislima:
On windows:
image

Also works the same as on macOS. If I stop and keep the files the game does not show as installed 🤔 
On the game card we still show the install button.
While downloading we show the cancel button (should have a pause though, will add it)

flavioislima:
There is no pause button on the game card on the current and next design so I will skip it.
I think makes sense since there will be the current download card on the new design.

BrettCleary:
Okay I found and fixed the issue with the game card showing as installed when paused. `downloadGame` was catching the error instead of throwing so `install` would go on to try and extract a canceled download. Now it rethrows correctly. this also fixes the notification showing installed vs just paused.

Thanks for fixing the .zip removal. 

As a note on windows when canceling a download and choosing to not keep downloaded files, not all of the files will be deleted because some of them are still in memory in hp client. When hp client is closed, all the files will be removed.

I think we should have subfolders of .temp with the appId to avoid naming collisions too

BrettCleary (approved):
I will leave it up to you whether we should put the .temp download chunks under an <app_id> subfolder on this branch or not

stormcakes (approved):

flavioislima:
Added the Pause button to the Card as well:
image
image

BrettCleary:
@flavioislima 
- yes I think it is fine that some of the chunks are still in memory. it is only around 40 MB or so
- sounds good re: pausing
- naming collisions could happen if games have the same title since the zip we download is the title I think right now, not the appid. the appid is the only property we know is unique
- nice will test the pause buttons now

BrettCleary:
I found an issue with the cancel button on the game page. folder was undefined so I could not cancel the download. I changed it to have the same logic as the game card. working now

BrettCleary:
Okay I did find one bug on windows
1. start a download
2. cancel the download and choose no, don't keep files. Some of the .temp files are still left due to an `Error: EPERM: operation not permitted, stat <chunk path here>`
3. try to download again and you receive the same eperm error because easydl tries to delete or overwrite one of the remaining chunks that is in memory still. so the download never starts

The only solution is to close out of the app (at which point the remaining chunks are deleted) and relaunch the app

I think this is a bug with easydl not releasing these files in memory before trying to delete them.

BrettCleary:
> Okay I did find one bug on windows
> 
> 1. start a download
> 2. cancel the download and choose no, don't keep files. Some of the .temp files are still left due to an `Error: EPERM: operation not permitted, stat <chunk path here>`
> 3. try to download again and you receive the same eperm error because easydl tries to delete or overwrite one of the remaining chunks that is in memory still. so the download never starts
> 
> The only solution is to close out of the app (at which point the remaining chunks are deleted) and relaunch the app
> 
> I think this is a bug with easydl not releasing these files in memory before trying to delete them.

This is the only issue I can replicate now on a clean install. I filed an issue here: https://github.com/andresusanto/easydl/issues/13 since I believe this to be an issue with easydl

stormcakes:
Retested after changes and no glaring issues atm. 
(22:31:25) INFO:    [DownloadManager]:  Violetear removed from download manager.

stormcakes (approved):

PR #290, opened 2023-04-28
[Fix] HyperPlay Store Manager

BrettCleary:
Fixes:
- uninstall folder name
- gameInfo version was not updating on refresh so updated games would always have the old version on install info
- ENOENT error when launching a game with an exe override for an exe that does not exist

To Test:
1. add another world to library
2. change library.json gameInfo.install.version to 0.1.21
3. refresh
4. game should be shown as updateable
5. update game
6. launch game

---

### DIFF
--- a/src/backend/storeManagers/storeManagerCommon/games.ts
+++ b/src/backend/storeManagers/storeManagerCommon/games.ts
@@ -3,7 +3,11 @@ import { GameConfig } from '../../game_config'
 import { isMac, isLinux, gamesConfigPath, icon } from '../../constants'
 import { logInfo, LogPrefix, logWarning } from '../../logger/logger'
 import path, { dirname, join, resolve } from 'path'
-import { appendFileSync, constants as FS_CONSTANTS } from 'graceful-fs'
+import {
+  appendFileSync,
+  existsSync,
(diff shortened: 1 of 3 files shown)


src/backend/storeManagers/storeManagerCommon/games.ts
```diff
@@ -160,7 +161,8 @@ export async function launchGame(
   const gameSettingsOverrides = await GameConfig.get(appName).getSettings()
   if (
     gameSettingsOverrides.targetExe !== undefined &&
-    gameSettingsOverrides.targetExe !== ''
+    gameSettingsOverrides.targetExe !== '' &&
+    fs.existsSync(gameSettingsOverrides.targetExe)
```
flavioislima, line 165: Instead of using this one from `fs`, import `existSync` from `graceful-fs` like we are doing on other methods. The graceful package seems more reliable than the default from node.

flavioislima (approved):
Good stuff!
One small comment only ;)

src/backend/storeManagers/storeManagerCommon/games.ts
BrettCleary: addressed in the latest commit

PR #296, opened 2023-05-03
[UI] Navigation and Library Update

BrettCleary:
WIP

---


flavioislima:
TEsting here and the UI looks pretty amazing and beautiful now.
Just noticed 2 things:
- No way to access Epic and GOG store
- Filter to show installed only not working

But looks promising! :D

flavioislima:
Now I can see the links to the stores, the popup is working as well (was some bug with my local build). and the Installed filter works fine now as well. Good stuff 👍🏽

flavioislima:
Only thing about the stores now is that, on the terminal, I can see this error when changing stores:

```
Error occurred in handler for 'GUEST_VIEW_MANAGER_CALL': Error:  (-3) loading 'https://store.epicgames.com/en-US/'
    at rejectAndCleanup (node:electron/js2c/browser_init:2:77333)
    at EventEmitter.failListener (node:electron/js2c/browser_init:2:77546)
    at EventEmitter.emit (node:events:525:35) {
  errno: -3,
  code: '',
  url: 'https://store.epicgames.com/en-US/'
}
Error occurred in handler for 'GUEST_VIEW_MANAGER_CALL': Error:  (-3) loading 'https://www.gog.com/'
    at rejectAndCleanup (node:electron/js2c/browser_init:2:77333)
    at EventEmitter.failListener (node:electron/js2c/browser_init:2:77546)
    at EventEmitter.emit (node:events:525:35) {
  errno: -3,
  code: '',
  url: 'https://www.gog.com/'
}
```

BrettCleary:
> Only thing about the stores now is that, on the terminal, I can see this error when changing stores:
> 
> ```
> Error occurred in handler for 'GUEST_VIEW_MANAGER_CALL': Error:  (-3) loading 'https://store.epicgames.com/en-US/'
>     at rejectAndCleanup (node:electron/js2c/browser_init:2:77333)
>     at EventEmitter.failListener (node:electron/js2c/browser_init:2:77546)
>     at EventEmitter.emit (node:events:525:35) {
>   errno: -3,
>   code: '',
>   url: 'https://store.epicgames.com/en-US/'
> }
> Error occurred in handler for 'GUEST_VIEW_MANAGER_CALL': Error:  (-3) loading 'https://www.gog.com/'
>     at rejectAndCleanup (node:electron/js2c/browser_init:2:77333)
>     at EventEmitter.failListener (node:electron/js2c/browser_init:2:77546)
>     at EventEmitter.emit (node:events:525:35) {
>   errno: -3,
>   code: '',
>   url: 'https://www.gog.com/'
> }
> ```

I can confirm this is also occurring in the main branch and in heroic when running in dev mode. 


These are the relevant electron snippets
https://github.com/electron/electron/blob/main/lib/browser/guest-view-manager.ts
https://github.com/electron/electron/blob/b02748e6076bd9f3802b7a4af1532d7e663381f5/lib/browser/navigation-controller.ts#L94-L104
From the second link:
```
// the webcontents has started another unrelated navigation in the
// main frame (probably from the app calling `loadURL` again); reject
// the promise
// We should only consider the request aborted if the "navigation" is
// actually navigating and not simply transitioning URL state in the
// current context.  E.g. pushState and `location.hash` changes are
// considered navigation events but are triggered with isSameDocument.
// We can ignore these to allow virtual routing on page load as long
// as the routing does not leave the document
return rejectAndCleanup(-3, 'ERR_ABORTED', url);
```

It seems to happen when one store page attempts to load before another one finishes, which is basically what the comment above says. This is fine though and doesn't seem to affect UX.

flavioislima:
Ok, tested this on the Steam deck and it's working pretty good. Need to double check the size of some elements in the screen with Billie but the navigation itself is working as it should.

I could both download and play a game. Of course the part of the metamask connection needs to be done with the virtual keyboard and it's not possible to navigate on the extension pop-up with the joystick which I think it's fine.

The stores does not work as well and this sis something we might need to investigate, perhaps on another pull request, since the hp store should be able to support the game pad navigation in my opinion.

Other than that everything else looks nice.

Good job on this one, Brett 😁

flavioislima (approved):

PR #317, opened 2023-05-10
[Feat] Auto Download Wine if no version was found in the system

flavioislima:
This will help new setups on mac and linux that does not have wine installed in their systems to play windows games.
If checks for any available wine on launch, if none was found, will download the recommended version: Wine-GE on Linux and Wine-Crossover on macOS. And then try to launch the game again.

---

### DIFF
--- a/public/locales/az/gamepage.json
+++ b/public/locales/az/gamepage.json
@@ -147,7 +147,7 @@
         "title": "Quraşdırılacaq komponentləri seçin"
     },
     "setting": {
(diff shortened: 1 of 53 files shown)


BrettCleary (approved):
This is working for me on steamdeck. After renaming the .proton .steam and .wine folders, I can see that the latest version of wine is installed and set as the default automatically. Another world launches correctly. Uninstalling this version of wine and reverting the folder names, I can see that the latest version of wine is not installed and proton is used as expected

PR #321, opened 2023-05-15
[FIX] Auto Download Wine on macOS

flavioislima:
Noticed this small bug after merging the PR.

With the current filter it won't find anything on macOS since it will try to find Wine-GE for it but it does not exists.

---


BrettCleary (approved):
LGTM. nice catch

PR #322, opened 2023-05-16
Support HyperPlay Store Games Hosted on Epic

BrettCleary:
For games with 
```
launch_epic?: boolean
epic_game_url?: string
```
defined, this [store PR](https://github.com/HyperPlay-Gaming/hyperplay-store/pull/11) will show the game page action button as "Play in Epic Store" and after clicking it, the user will be navigated to the corresponding listing in the Epic Store.

When the game is added and getGameInfo is called for the epic game, a map (generated on client start) will merge the HyperPlay listing `GameInfo` metadata with the legendary metadata. This way we can still show web3/HyperPlay specific metadata like token requirements even though the game is managed by legendary

To test, you will need to 
1. replace the developers.hyperplay.api/listings axios calls with a test.json that you import.
2. manually change one of the game's metadata to have `launch_epic: true` and `epic_game_url: https://store.epicgames.com/en-US/p/fall-guys` or some other epic game that you have added to your library
3. run [this store PR branch](https://github.com/HyperPlay-Gaming/hyperplay-store/pull/11)
4. set the store url in hp client to http://localhost:3000
5. navigate to the hp store in the client
6. navigate to the game you modified in the hp store
7. Click "Play in Epic Store"
8. See that it navigates you to the epic store. once you do that make sure the game is added to your legendary/epic library and go to the library screen
9. Click on the epic games listing and you should see some of the fields from the hp listing like "has web3 features: true"

---

### DIFF
--- a/src/backend/storeManagers/legendary/library.ts
+++ b/src/backend/storeManagers/legendary/library.ts
@@ -618,7 +618,7 @@ async function loadAll(): Promise<string[]> {
   if (existsSync(legendaryMetadata)) {
     const loadedFiles: string[] = []
     for (const appName of allGames) {
-      const wasLoaded = await loadFile(appName + '.json')
+      const wasLoaded = loadFile(appName + '.json')
       if (wasLoaded) {
         loadedFiles.push(appName)
       }
--- a/src/backend/storeManagers/legendary/library.ts
+++ b/src/backend/storeManagers/legendary/library.ts
@@ -41,6 +41,7 @@ import { callRunner } from '../../launcher'
 import { dirname, join } from 'path'
 import { isOnline } from '../../online_monitor'
 import { update } from './games'
+import { epicTitleToHpGameInfoMap } from '../hyperplay/utils'
 
 const allGames: Set<string> = new Set()
 let installedGames: Map<string, InstalledJsonMetadata> = new Map()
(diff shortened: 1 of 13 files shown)


src/backend/storeManagers/legendary/library.ts
```diff
@@ -584,12 +617,12 @@ function loadFile(fileName: string): boolean {
 async function loadAll(): Promise<string[]> {
   if (existsSync(legendaryMetadata)) {
     const loadedFiles: string[] = []
-    allGames.forEach((appName) => {
-      const wasLoaded = loadFile(appName + '.json')
+    for (const appName of allGames) {
+      const wasLoaded = await loadFile(appName + '.json')
```
flavioislima, line 621: Can you elaborate more on this change?
Not sure if this can impact or not big libraries loading.

flavioislima (approved):
Looks good. Left a small comment 👍🏽

src/backend/storeManagers/legendary/library.ts
BrettCleary: ah yeah so loadFile was originally synchronous. I made it async to fetch hp metadata inside it but then decided it would be better to cache it as a map so I then reverted it back to sync. The await does nothing here. I will remove it

I switched from forEach to for of because forEach does not wait for promises to resolve but I can await the promise in the for of loop

performance here is identical to before I believe

PR #328, opened 2023-05-19
[Fix] Installation of HP games, statuses, filters, stuck download and more

flavioislima:
- Fix games like Altiros failing to install due to unzipping errors for files with non-ASCII characters.
- Fix Progress report time, using the throttle method to report once every 1 second instead of ms.
- Proper calculate ETA and send to frontend
- Properly calculate the disk write speed.
- Fix some places showing raw bytes instead of MB.
- Set concurrent connection to 1 so the downloads will start right away, this makes the downloads slower though.
- Fix Default platform selection for HP games on install
- UI: show extracting messages and block the install button on the game page.
- Implemented HP `getExtraInfo` so we can eliminate warning messages due to it not being implemented.
- Some improvements to logging when downloading HP Games.
- Fixed Operating System library filter not working for HP games
- Fixed OS and Install size information on the game page

---


BrettCleary (approved):
added some unit tests and fixed an issue with `extractZip`
ran through manual tests for all the FE stuff.
looks good. I will go ahead and approve/merge since this will be shipped in the hotfix

PR #330, opened 2023-05-23
Add e2e testing

BrettCleary:
Adds e2e testing with playwright

---


stormcakes:
Taking a look currently. Running into issues trying to get the api.spec.ts to appear in the "Testing" plugin but could just be how it was set-up. Looks like tests are running in "Test Results" though so looks good! Would also add the playwright-report folder to .gitignore. Results could be sent via the CLI so the report folder wouldn't need to be persistent :)

stormcakes (approved):
Tested and looks buenooo :)

PR #344, opened 2023-06-01
[Feature] Add copy base uri button

BrettCleary:
This allows the user to use wallet connect with wallets that cannot parse out the base uri from the metamask deep link qr code

To test Ledger Live for example
1. Go to accounts
2. click on an account
3. click wallet connect icon in top right
4. paste URL you copied with the copy base uri button

#### Screenshots
wallet connect has base uri button
image

mm mobile does not have the base uri button
image

---

### DIFF
--- a/src/backend/utils.ts
+++ b/src/backend/utils.ts
@@ -44,7 +44,6 @@ import {
 import { promisify } from 'util'
 import i18next, { t } from 'i18next'
 import si from 'systeminformation'
-import unzipper from 'unzipper'
 
 import {
   fixAsarPath,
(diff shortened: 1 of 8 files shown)


stormcakes (approved):
Worked perfectly for my Ledger :)

src/backend/utils.ts
```diff
@@ -86,7 +85,6 @@ import {
   updateWineVersionInfos,
   wineDownloaderInfoStore
 } from './wine/manager/utils'
-import { clean } from 'easydl/dist/utils'
```
flavioislima, line 89: Is this change related? also the translations update? or it came from another branch?

flavioislima (approved):
I think looks fine, but the branch has some unrelated changes I guess. 👍🏽

src/backend/utils.ts
BrettCleary: These variables were unused. Not sure how they got back into develop after your replaced unzipper with the other library.

PR #369, opened 2023-06-18
[Fix] Uninstall games list layout shift

BrettCleary:
Fixes empty game card sized div when uninstall modal is shown

---


flavioislima (approved):
Looks good. 👍🏽 
Another one-line solution is to add `position: absolute` to the `.Dialog` class.

stormcakes (approved):

BrettCleary:
> Looks good. 👍🏽 Another one-line solution is to add `position: absolute` to the `.Dialog` class.

yes. I don't think we need the extra dialog div element though

PR #375, opened 2023-06-20
Merge Develop

BrettCleary:
Merge then delete develop branch

---


BrettCleary:
tested with moonblasters. sub bullet points are notes regarding the test. the checkmark indicates that it has been tested and is working

### HP
- [X] download, install launch
    - sometimes it finishes download at 95-99% and doesn't send the 100% event to the frontend so it stays at 95-99% while it is extracting
- [X] download pause resume
- [X] download pause cancel
- [X] download cancel don't keep (requires app restart to redownload)
- [X] download cancel and keep, then download again
    - there was an intermittent issue with the install path on the download again step installation modal showing as default. this has been fixed now, using https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/commit/a1266eb780b302075ef61c0f7777a2604c5d8ef2 as a reference where flavio fixed previously
- [X] stop playing game

BrettCleary:
for epic, tested with hundred days - wine simulator and fortnite

- [X] Download, launch
- [X] Download, pause, resume, launch
    - i noticed that on resume the % starts about 10-15% behind where it was when it was paused
    - I also paused right at 100% downloaded and now I cannot install again or uninstall. the game shows as uninstalled. this seems to be a legendary issue. it persists on relaunching HP. after running `legendary cleanup` I can reinstall. I think we should add this command to a settings item labeled "reset epic store manager" or something like that
    - also on uninstall and starting the next install. the progress bar is initially at 100% on the game card from the last download until the next progress update is sent to the frontend. we should send a progress event from the backend when a new download starts that has 0%
- [X] download pause cancel
- [X] download cancel don't keep, download again
- [X] download cancel and keep, then download again
- [X] stop playing
    - noticed the stop button shows, then disappears when launching the game. this is probably due to it launching a process that launches the game process then quits. not sure if we can handle this for non HP games though since we need the game's process name to track the process. this is the same behavior as main I believe

flavioislima:
I think you need to skip signing on Windows for PRs, like we do for macOS, otherwise it won't pass since it cannot find the certificate.

Kwilke1982:
> Found one small regression on the Add Game screen. This was fixed some time ago on this PR: #328
> 
> It should show only the platforms for that Operating system since it is not possible to play Linux games on macOS and vice versa. image
> 
> the logic is here on this file:
> 
> https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/blob/11ed75ac7df5ec1ff4477d82d43d6f3ee14c9e36/src/frontend/screens/Library/components/InstallModal/index.tsx#L74

We should still have the option for Mac users to select windows native or mac os builds when installing games that offer both.

flavioislima:
> We should still have the option for Mac users to select windows native or mac os builds when installing games that offer both.

Yes, this is only for the Add game screen. Although might be affecting the Download screen as well since it uses the same logic.

BrettCleary:
> Found one small regression on the Add Game screen. This was fixed some time ago on this PR: #328
> 
> It should show only the platforms for that Operating system since it is not possible to play Linux games on macOS and vice versa. image
> 
> the logic is here on this file:
> 
> https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/blob/11ed75ac7df5ec1ff4477d82d43d6f3ee14c9e36/src/frontend/screens/Library/components/InstallModal/index.tsx#L74

this is the same behavior on main and the code from the pr you linked is identical to what is on this develop branch. we can refactor/fix this after merging

flavioislima:
> > Found one small regression on the Add Game screen. This was fixed some time ago on this PR: #328
> > 
> > It should show only the platforms for that Operating system since it is not possible to play Linux games on macOS and vice versa. image
> > 
> > the logic is here on this file:
> > 
> > https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/blob/11ed75ac7df5ec1ff4477d82d43d6f3ee14c9e36/src/frontend/screens/Library/components/InstallModal/index.tsx#L74
> 
> this is the same behavior on main and the code from the pr you linked is identical to what is on this develop branch. we can refactor/fix this after merging

Have you checked 0.2.3?

flavioislima (approved):
I think there are still minor bugs but I think we can fix them later. Most of it looks good 👍

stormcakes (approved):
Flagged a few minor bugs but major ones were resolved. Looks good :)

BrettCleary:
> Have you checked 0.2.3?

yes it is same behavior as main

flavioislima:
> > Have you checked 0.2.3?
> 
> yes it is same behavior as main

Dafuq, will test again tomorrow 👍

PR #386, opened 2023-06-23
[Tech] Compile main to bytecode

BrettCleary:
Compile main to bytecode

---


flavioislima:
I am getting Segmentation fault when trying to run on macOS and then I get this screen with the error:
image

reading the docs, it says that we need to disable the sandbox feature on electron for this to work, this might be the reason. Probably other changes might be needed as well.

BrettCleary:
> reading the docs, it says that we need to disable the sandbox feature on electron for this to work, this might be the reason. Probably other changes might be needed as well.

this is only necessary for preload scripts because it imports vm from node in the preload script so nodeIntegration is required and sandbox has to be false. But we are not compiling any preload scripts to bytecode

flavioislima:
> > reading the docs, it says that we need to disable the sandbox feature on electron for this to work, this might be the reason. Probably other changes might be needed as well.
> 
> this is only necessary for preload scripts because it imports vm from node in the preload script so nodeIntegration is required and sandbox has to be false. But we are not compiling any preload scripts to bytecode

I see, so we need to get deeper in the docs to see why is not working then 🤔

BrettCleary:
I can verify this is working on windows and Ubuntu 22.04.2 LTS

The flatpak build was missing dependencies when compiling the byteloader. I've added them to the gh action and now it is building

BrettCleary:
> I am getting Segmentation fault when trying to run on macOS and then I get this screen with the error: image
> 
> reading the docs, it says that we need to disable the sandbox feature on electron for this to work, this might be the reason. Probably other changes might be needed as well.

i can replicate this on m1 also. will look into it

BrettCleary:
i've tested and verified with
- flatpak
- appimage
- mac m1m2 dmg 
- windows setup

all are working and main process was verified to be compiled to bytecode so I will merge as this is blocking other work

PR #393, opened 2023-06-29
[Fix] Linux missing app icon

flavioislima:
Fix missed app icon on linux. since the builds were failing due to icns icon on linux, I replaced it with the png and png does not work on linux. It needs to be ICNS for some weird reason.

So I am reverting this change.

---


BrettCleary (approved):
Icon looks good on linux and windows!

PR #398, opened 2023-07-04
[Feat] cross platform overlay

BrettCleary:
cross platform overlay

### Linux
image
image

### Windows
image
image

---


flavioislima:
macOS:
Screenshot 2023-07-07 at 22 50 23
Screenshot 2023-07-07 at 22 52 47


Oni's Quest still doesn't work unless it changes to the borderless fullscreen. (haven't tested though)
Sometimes the toast appears before the game starts, so maybe we could increase the delay a bit.

Could not test any windows game on mac because wine is totally broken on Sonoma Beta, but I tested before and worked fine and should work as well since I could see that it creates the overlay on top of all macOS apps on all workspaces.

flavioislima (approved):
That's pretty great job, man. There are a few things to improve but this looks great already. :)

BrettCleary:
> Oni's Quest still doesn't work unless it changes to the borderless fullscreen. (haven't tested though)
> Sometimes the toast appears before the game starts, so maybe we could increase the delay a bit.
> 
> Could not test any windows game on mac because wine is totally broken on Sonoma Beta, but I tested before and worked fine and should work as well since I could see that it creates the overlay on top of all macOS apps on all workspaces.
> There is a small space on the bottom always, like it reserves space for the dock, not sure how to fix it.


thx for testing
- I just changed the delay from 3 seconds to 5 seconds
- sounds good. i just retested on windows and it is working well
- hmm not sure how to fix either. let's fix on separate branch. i opened an issue here so we don't forget https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/issues/406

PR #401, opened 2023-07-06
[Feat] Browser game fullscreen hotkey

BrettCleary:
toggles fullscreen in browser games with F11 hotkey like regular browsers

---

### DIFF
--- a/src/backend/main_window.ts
+++ b/src/backend/main_window.ts
@@ -78,7 +78,7 @@ export const createMainWindow = () => {
     }
   })
 
-  mainWindow.webContents.on('before-input-event', (event, input) => {
+  mainWindow.webContents?.on('before-input-event', (event, input) => {
     if (input.key === 'F11') backendEvents.emit('toggleFullscreen')
   })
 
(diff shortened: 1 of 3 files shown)


src/backend/main.ts
```diff
@@ -358,6 +358,8 @@ if (!gotTheLock) {
       OverlayApp.toggleIntercept
     )
 
+    globalShortcut.register('F11', () => backendEvents.emit('toggleFullscreen'))
```
flavioislima, line 361: I wonder if we could just use a local shortcut for that that is created after the browser game was created. Although F11 should be fine, using GlobalShortcut will override all F11 shortcuts on all apps that uses it while HyperPlay is opened. That might not be a big deal though, but a local shortcut could work better.

flavioislima (approved):
Left one small comment but should not block the merge, just something to look into later.

stormcakes (approved):

src/backend/main.ts
BrettCleary: great idea. i didn't think about this. just changed it

PR #416, opened 2023-07-26
[Tech] Add LavaMoat allow scripts

BrettCleary:
see https://github.com/LavaMoat/LavaMoat/tree/main/packages/allow-scripts for more info on the package

Basically we disable all the npm package lifecycle scripts by default and only allow the ones listed in the package.json through.

Instead of installing dependencies with `yarn` or `yarn install`, we will now have to use `yarn setup` or the allowed package lifecycle scripts will not be executed

The fortawesome packages just log their license in the postinstall script so these can be disabled

@lavamoat/preinstall-always-fail is a package that just throws on preinstall to check that we are ignoring scripts correctly

`discord-rich-presence-typescript>discord-rpc>register-scheme` is not necessary since it can import electron
- see https://github.com/discordjs/RPC/blob/9e7de2a6d917591f10a66389e62e1dc053c04fec/src/util.js#L4-L11

I've checked the other packages and they need their lifecycle scripts to install correctly

---

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,9 +35,9 @@ jobs:
       - name: Authenticate with private NPM package
         run: echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" > ~/.npmrc
       - name: Install modules.
-        run: npm run setup:npm
+        run: yarn setup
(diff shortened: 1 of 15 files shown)


flavioislima (approved):
LGTM

PR #417, opened 2023-07-27
[Fix] Sentry init after ready

BrettCleary:
Metrics fail on initial opt in due to sentry init throwing from being called after app.whenReady() 

This removes the failing sentry init. This also means we won't get error logs on initial launch if the user opts in to metrics. We should consider an upstream PR to https://github.com/getsentry/sentry-electron to add an enable/disable toggle to sentry after init is called 

error message: 
```
Error occurred in handler for 'changeMetricsOptInStatus': SentryError: Sentry SDK should be initialized before the Electron app 'ready' event is fired
```

---

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -9,8 +9,7 @@ import {
   GamepadInputEvent,
   WineCommandArgs,
   ExecResult,
-  Runner,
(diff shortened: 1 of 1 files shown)


flavioislima (approved):
LGTM

PR #425, opened 2023-07-28
[Fix] Browser game fullscreen toggle

BrettCleary:
Fixes toggle of browser game fullscreen with F11

---

### DIFF
--- a/src/backend/main_window.ts
+++ b/src/backend/main_window.ts
@@ -1,7 +1,6 @@
 import { BrowserWindow, screen, app } from 'electron'
 import path from 'path'
 import { configStore } from './constants'
-import { backendEvents } from './backend_events'
 
 let mainWindow: BrowserWindow | null = null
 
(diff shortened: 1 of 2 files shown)


flavioislima:
Tested here and on Linux with KDE it doesn't do anything.
I mean, it shows like it would leave fullscreen, the window move a bit and then it gets back to fullscreen. Only way to leave the game is by using ALT + F4

flavioislima:
> > Tested here and on Linux with KDE it doesn't do anything. I mean, it shows like it would leave fullscreen, the window move a bit and then it gets back to fullscreen. Only way to leave the game is by using ALT + F4
> 
> This is because in dev mode, electron already provides a keybind of F11 to toggle fullscreen, so it is called twice and gives the behavior you described. You can also see this because the main window can be toggled fullscreen in dev mode but not in the packaged build.
> 
> In a packaged build, F11 will toggle correctly for mighty action heroes. I did just notice that DFK is not toggling.
> 
> I added `event.preventDefault()`. This allows F11 to have the same behavior in dev mode and also fixes the DFK issue

I didn't know that was a difference between dev mode and the build on this matter.
Tested now with the build and it works fine 👍🏽

flavioislima (approved):
LGTM

nyghtstalker (approved):

PR #426, opened 2023-07-28
[Fix] Linux Overlay

BrettCleary:
Fixes the linux native overlay

---


flavioislima:
Tested on Linux with the Kosium native game. The overlay works fine.
One thing that I believe could be improved is that it could wait a bit more to show.
in my case, the overlay appears first and then the game does not shows in the foreground.
So I can see the MM extension and the toast. Then I need to click on the game from the taskbar for it to go to the foreground. Then I can trigger the overlay again and it works nicely on top of the game.

So perhaps if we could wait a few more seconds to show it the experience could be better.

BrettCleary:
> Tested on Linux with the Kosium native game. The overlay works fine. One thing that I believe could be improved is that it could wait a bit more to show. in my case, the overlay appears first and then the game does not shows in the foreground. So I can see the MM extension and the toast. Then I need to click on the game from the taskbar for it to go to the foreground. Then I can trigger the overlay again and it works nicely on top of the game.
> 
> So perhaps if we could wait a few more seconds to show it the experience could be better.

added another 2.5 seconds and opened a hp overlay submodule PR for multiple screen fix and one other fix detailed in the hp overlay submodule PR

flavioislima (approved):
It works pretty good on my end now.
we now just need to figure it out how to deal with it on the Steam deck, because on gaming mode the game is launched and then the black screen stays on top and we cannot see the game. There is also no way to trigger the overlay on and off using the steam controller. But this can be solved on another PR ofc.

PR #463, opened 2023-08-15
[Metrics] Ensure unique metrics id on launch

BrettCleary:
Checks that an opted in user has a unique metrics id on app launch

Reports a 'Metrics Error Correction' event if the user is opted in but still has the anonymous id so that we will know how many people were affected historically

---

### DIFF
--- a/src/backend/metrics/metrics.ts
+++ b/src/backend/metrics/metrics.ts
@@ -189,7 +189,9 @@ export const trackScreen = async (name: string, properties?: apiObject) => {
 function ensureUniqueMetricsId() {
   const currentId = metricsStore.get('metricsId')
   const status = metricsStore.get('metricsOptInStatus')
-  if (currentId === ANONYMOUS_ID && status === MetricsOptInStatus.optedIn) {
+  const isCorrectId = /0x[0-9,a-f, A-F]{48}/
+
+  if (!isCorrectId.test(currentId) && status === MetricsOptInStatus.optedIn) {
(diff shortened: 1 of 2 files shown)


src/backend/metrics/metrics.ts
```diff
@@ -181,6 +181,24 @@ export const trackScreen = async (name: string, properties?: apiObject) => {
...
+  const status = metricsStore.get('metricsOptInStatus')
+  if (currentId === ANONYMOUS_ID && status === MetricsOptInStatus.optedIn) {
+    metricsStore.set('metricsId', generateRandomId())
+    trackEvent({
+      event: 'Metrics Error Correction'
+    })
+  }
+}
+
+ensureUniqueMetricsId()
```
eliobricenov, line 200: Why this is called right after the definition instead of a main process?

eliobricenov (approved):
LGTM. Only one question

src/backend/metrics/metrics.ts
BrettCleary: it will be called when the module is first imported before the app calls any of the track methods

BrettCleary: so electron will start running the main.ts module, everything will be imported and top level stuff like this run in those imported modules, then main will create the windows, launch the app etc.

nyghtstalker (approved):
After changing ID to 0s I confirmed that launching the game will restore to the previous value.

PR #484, opened 2023-08-20
[CI/CD] Check MetaMask is latest in GHA 

BrettCleary:
Check that MetaMask is on the latest release in a github workflow

See the relevant issue for more details https://github.com/HyperPlay-Gaming/product-management/issues/22

---

### DIFF
--- a/.github/workflows/checkMetaMaskLatest.yml
+++ b/.github/workflows/checkMetaMaskLatest.yml
@@ -1,0 +1,33 @@
+name: Check MetaMask is up to date
+
+on:
+  pull_request:
+    branches: [main]
+
+env:
+  GITHUB_TOKEN: ${{ secrets.WORKFLOW_TOKEN }}
+  GH_TOKEN: ${{ secrets.WORKFLOW_TOKEN }}
+
+jobs:
+  draft-releases:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository.
+        uses: actions/checkout@v3
+        with:
+          submodules: 'recursive'
+          token: ${{ secrets.pat }}
+      - name: Setup NodeJS
+        uses: actions/setup-node@v3
+        with:
+          node-version: '18'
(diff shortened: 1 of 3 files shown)


cicd/checkMetaMaskIsLatest.ts
```diff
@@ -0,0 +1,37 @@
...
+
+async function main() {
+  const metaMaskReleases = child_process.spawnSync('gh', [
+    'release',
+    'list',
+    '-R',
+    'MetaMask/metamask-extension'
+  ])
+  const metaMaskReleasesStdOut = metaMaskReleases.stdout?.toString()
+  const latestVersion = metaMaskReleasesStdOut.match(
```
flavioislima, line 12: What do you think of using this package here to check for the versions? 
https://www.npmjs.com/package/semver

It probably will cover more cases than doing it manually. We use it on the client as well and has only one dependency.

.github/workflows/checkMetaMaskLatest.yml
```diff
@@ -0,0 +1,33 @@
+name: Check MetaMask is up to date
+
+on:
+  pull_request:
```
flavioislima, line 4: Will this be required to pass before merging a PR?
Because I believe this will fail from time to time in our PRs and we won't update it on these pull requests.

flavioislima (approved):
Looks good. Left some small comments. 👍

eliobricenov (approved):
Same comments from Flavio

cicd/checkMetaMaskIsLatest.ts
BrettCleary: yeah now that you bring it up, I think we should honestly just check if we're on the same minor version as MetaMask. The patches are updated ~1 per week but the minor versions are ~1 per month.

BrettCleary:
closed in favor of a husky pre-push tag command

PR #489, opened 2023-08-25
refactor(webview): use keys in webview components routes

eliobricenov:
Here's a good explanation of the current behavior https://www.loom.com/share/b76e1dac916242788a2e81fa51df0a01 please watch first

By using different React keys, we can ensure that the webview is unmounted on navigation. See https://www.makeuseof.com/react-keys-avoid-component-conflict/

**The caveats with this approach are two, both cause by the webview being unmounted:**

1. The history is lost between navigation. If you go from one route that uses webview to another, you cannot use the webview's navigation go back arrow to go to the previous page.
2. Going to a different route makes the compontent to reload the page even if it has visited before. In the current implementation because the webview component is never unmounted, it seems the webview caches the urls.

### DIFF
--- a/src/frontend/App.tsx
+++ b/src/frontend/App.tsx
@@ -59,10 +59,13 @@ function App() {
             />
             <Route path="/library" element={<Library />} />
             <Route path="login" element={<Login />} />
-            <Route path="hyperplaystore" element={<WebView />} />
-            <Route path="epicstore" element={<WebView />} />
-            <Route path="gogstore" element={<WebView />} />
-            <Route path="wiki" element={<WebView />} />
+            <Route
+              path="hyperplaystore"
+              element={<WebView key="hyperplaystore" />}
+            />
+            <Route path="epicstore" element={<WebView key="epicstore" />} />
+            <Route path="gogstore" element={<WebView key="gogstore" />} />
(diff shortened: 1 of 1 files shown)


BrettCleary:
what if instead of separate routes for all the stores, we just had one route and used a mobx store to change the webview url using its ref.

I think this way we could keep the history and prevent the issue we're fixing here (i.e. where you go back but then can't click the route you were just on because you're still on it).

I don't quite understand your point # 2. Would this implementation address that as well?

eliobricenov:
> what if instead of separate routes for all the stores, we just had one route and used a mobx store to change the webview url using its ref.
> 
> I think this way we could keep the history and prevent the issue we're fixing here (i.e. where you go back but then can't click the route you were just on because you're still on it).
> 
> I don't quite understand your point # 2. Would this implementation address that as well?

Yes, if we don't unmount the component we can keep the history. What I mean by point # 2 is that when you mount/unmount you see the "Loading website" message.

BrettCleary (approved):
LGTM

nyghtstalker:
Navigation issues are corrected now that back button on Epic and GOG are disabled.

nyghtstalker (approved):

Issue #439, opened 2023-08-05
Update filter position


BrettCleary:
### Describe the bug

The position on this filter type is also different from the one we have on Figma:

![image](https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/assets/27568879/7f2a5800-4bc5-4a76-a20e-c680635c86ff)

Figma version: ![image](https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/assets/27568879/69464ae0-a967-4bcb-82f6-08bd92bdde7a)

### Add logs

.

### Steps to reproduce

.

### Expected behavior

.

### hyperplay Version

main branch from GitHub

### System Information

.

Issue #93, opened 2023-01-09
Library design for table view


pimentel08:
As a gamer, I want to be able see all the games I have downloaded from the HyperPlay store, Epic store, and GoG in a single table view along with their status, if not fully installed, so that I can have an easy way of viewing and managing my large library of games. 

**Description**
Using new UI library, make updates to table with states such as:

- Uninstalled
- Installing
- Cancel

**AC**
- New Library view using the attached designs. 
- Ability to sort view by game store
- View status of games: 
    - Installed
    - Installing 
    - Uninstalled 

- Action button that allows a user to play, install, cancel installation, or remove game.  

[Figma Link](https://www.figma.com/file/GKPELZxk551oVC3EDLgEDA/Game-Launcher?node-id=1372%3A12483&t=ALoigxEcCpDzVoup-4)

wswarren12:
Should non-installed games show up in the table? Or is that to capture when a game is in the Queue (in which case I'd put its status as Queued).

biliesilva:
@wswarren12 This screen is the same screen as Library, but with a table view (the other is a card view). So it will show all the games in the Library, installed or not, but with the possibility for the user to filter what he wants at the moment.

wswarren12:
Thanks for the clarification @biliesilva. Just to make sure I'm not missing anything...

The idea is that, as a user, if I'm in my Library and switch to the table view, I'll see the following categories of games:

-Downloaded Games
-Downloads In-Queue
-Games from the HyperPlay store that haven't been downloaded yet?

But we'll add a filter, so that I can also filter for just already downloaded games or games waiting to complete downloads.

biliesilva:
Welcome @wswarren12 

In the Library, you will see the games you have from Epic and GOG accounts (Installed or not). And for the HyperPlay games, you will see in the Library the games you have added (browser games) or downloaded from HyperPlay Store. It will be the same as the grid view.

wswarren12:
Ok perfect, thanks for clarifying @biliesilva.

pimentel08:
https://www.figma.com/file/GKPELZxk551oVC3EDLgEDA/HP-Launcher?node-id=1372-12483&t=t3BwcL4Mja9QMHKW-4

pimentel08:
[...]

PR #500, opened 2023-08-30
[UI] Update Filter Position, Refactor Library Top Bar

BrettCleary:
closes https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/issues/439

Also refactors the library top bar into its own component since the Library component is getting pretty big at ~700 lines previously.

### Screenshots
image
image
image

---


src/frontend/screens/Library/components/LibraryTopBar/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+                {otherFiltersData.map((val, index) => (
+                  <Menu.Item
+                    closeMenuOnClick={false}
+                    key={`toggleItem${index}`}
+                  >
+                    <Toggle
+                      defaultChecked={val.defaultValue}
+                      labelPosition="right"
+                      onChange={
+                        //eslint-disable-next-line
```
eliobricenov, line 93: this can be avoided with https://github.com/HyperPlay-Gaming/hyperplay-ui/pull/63

src/frontend/screens/Library/components/LibraryTopBar/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+                selected={selectedFilter}
+                targetWidth={275}
+              />
+            </div>
+            <div>
+              <GenericDropdown
+                target={
+                  <GenericDropdown.GenericButton
+                    text={'Other filters'}
+                    style={{ width: '340px' }}
```
eliobricenov, line 80: I know this is an old component but can we use a css module instead of inline styles now that we're changing it?

src/frontend/screens/Library/components/LibraryTopBar/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+                      >
+                        {val.text}
+                      </div>
+                    </Toggle>
+                  </Menu.Item>
+                ))}
+              </GenericDropdown>
+            </div>
+            <div id="alignEnd">
+              {/* <div>
```
eliobricenov, line 114: We can remove this too

src/frontend/screens/Library/index.tsx
```diff
@@ -552,90 +553,7 @@ export default React.memo(function Library(): JSX.Element {
...
-            </div>
-            <div>
-              <Button type="tertiary" className={styles.gridListButton}>
-                <Images.List fill="white" height={24} width={24} />
-              </Button>
-            </div> */}
-            </div>
-          </Tabs.List>
-        </Tabs>
+        <LibraryTopBar handleCategory={handleCategory} category={category} filters={filters} setSelectedFilter={setSelectedFilter} selectedFilter={selectedFilter} otherFiltersData={otherFiltersData} isEpicLoggedin={isEpicLoggedin} isGOGLoggedin={isGOGLoggedin}/>
```
eliobricenov, line 556: `isEpicLoggedin`, `isGOGLoggedin` can be calculated in 
 `<LibraryTopBar/>` via `useContext(ContextProvider)`.

`category` and `handleCategory` are available through the context too.

Overall, I think that you can use a context or a store to handle the state and avoid prop drilling for this refactor

eliobricenov: I appreciate refactor initiative btw 🔥

src/frontend/screens/Library/components/LibraryTopBar/index.tsx
BrettCleary: sounds good

src/frontend/screens/Library/components/LibraryTopBar/index.tsx
BrettCleary: we actually still need to implement this feature and these icons are set up perfectly. I'd rather just leave them https://www.figma.com/file/GKPELZxk551oVC3EDLgEDA/HP-Launcher?type=design&node-id=3508-25495&mode=design&t=JSbrhwRJhQ2LcoyW-4

BrettCleary: here's the issue: https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/issues/93 

it's out of scope for this pr though

src/frontend/screens/Library/index.tsx
BrettCleary: tru. after the refactor there are only a few props though and it is only one level. i think it is fine to use props for that

src/frontend/screens/Library/components/LibraryTopBar/index.tsx
BrettCleary: sure

flavioislima:
@BrettCleary can you add some screenshots/videos to the PR so we can see the changes? :)

package.json
```diff
@@ -185,6 +186,8 @@
     "easydl": "^1.0.3",
     "electron-store": "^8.0.1",
     "electron-updater": "^5.0.1",
+    "embla-carousel-autoplay": "^8.0.0-rc12",
+    "embla-carousel-react": "^8.0.0-rc12",
```
flavioislima, line 190: What is this library for? I cannot see it being imported in the code anywhere.

BrettCleary: It is a peer dependency from the ui library. We cannot ship it as a dependency currently because it breaks nextjs. This is because mantine loses the isBrowser/isNode checks when being built in the ui library

When updating to the latest ui library version for elio's fix for the generic button on change typing, i also had to add these as it included the carousel refactor

eliobricenov (approved):
LGTM

package.json
flavioislima: That's not really convenient. So on every place we need to use the UI we also need to install these dependencies? What about trying a different way of bundling the UI lib?

flavioislima (approved):

package.json
BrettCleary: yes. I have spent a decent amount of time on this and it is pretty complex/involved. Not worth the time at the moment in my opinion. we only use the ui library in a couple of repos

PR #505, opened 2023-09-01
[CI] Windows Release and Signing

flavioislima:
This sets a Windows release workflow that will be triggered with the others and automatically sends the binary to be signed on signpath and then gets back the artifacts and updates the latest.yml with the new information and creates or updates the release.

---


yarn.lock
BrettCleary, line 1: This is a lot of updates. Is this from `yarn upgrade`? 

Looks like these dependency updates are causing the builds to fail too. 

Maybe we should revert to the previous lockfile and handle these updates on a separate branch

BrettCleary (changes requested):
GHA looks good. All the builds are failing now though. I think it's due to the yarn.lock updates so maybe we should revert those

yarn.lock
flavioislima: Fixed

BrettCleary (approved):
LGTM

eliobricenov (approved):

PR #513, opened 2023-09-06
[CI] Fix utils test rounding error

BrettCleary:
This utils test sometimes fails due to small delays between when
`const lastProgressTime = Date.now().valueOf() - 1000 * 4` 
is called in the test and when 
`const elapsedSeconds = (Date.now() - lastProgressTime) / 1000`
is called in `calculateEta`

We should just accept 04 or 05 to pass the test

---

### DIFF
--- a/src/backend/__tests__/utils.test.ts
+++ b/src/backend/__tests__/utils.test.ts
@@ -198,7 +198,7 @@ describe('backend/utils.ts', () => {
         getBytesFromMB(totalSizeMB),
         lastProgressTime
       )
-      expect(etaString).toEqual('00:00:04')
+      expect(etaString).toEqual(expect.stringMatching(/00:00:0[4,5]/))
     })
 
     test('normal download minutes', () => {
--- a/src/backend/__tests__/utils.test.ts
+++ b/src/backend/__tests__/utils.test.ts
@@ -172,7 +172,7 @@ describe('backend/utils.ts', () => {
         getBytesFromMB(downloadSpeedMB),
         getBytesFromMB(totalSizeMB)
       )
-      expect(etaString).toEqual('00:00:09')
+      expect(etaString).toEqual(expect.stringMatching(/00:00:0[8,9]/))
     })
 
     test('downloaded > total size', () => {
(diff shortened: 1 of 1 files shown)


eliobricenov (approved):
lgtm

src/backend/__tests__/utils.test.ts
```diff
@@ -198,7 +198,7 @@ describe('backend/utils.ts', () => {
         getBytesFromMB(totalSizeMB),
         lastProgressTime
       )
-      expect(etaString).toEqual('00:00:04')
+      expect(etaString).toEqual(expect.stringMatching(/00:00:0[4,5]/))
```
eliobricenov, line 201: maybe we should make this more general?
```js
expect(etaString).toEqual(expect.stringMatching(/00:00:0\d/))
```

BrettCleary: I haven't seen it more than 1 second off. Theoretically if there's no delay, it should be 05, but with the delay it's 04. If we did `0\d` then it could handle being up to 5 seconds off, but I think if we have >1 second off, it's more than likely a problem with the function we're testing

I'd recommend just matching 04 and 05 and if we have another failure at 03 where it is a timing issue and not a bug with calculateEta, then we change it to match 03-05

eliobricenov: Roger that! Thanks for the explanation 😄

flavioislima (approved):
LGTM
but I am still not sure how a test like this, that is never exact and needs this type of check,  really brings something to the client.

BrettCleary:
> LGTM but I am still not sure how a test like this, that is never exact and needs this type of check, really brings something to the client.

We should have unit tests for all of our pure functions in my opinion. `calculateEta` just happens to have this timing issue due to the nature of the function and javascript testing.

BrettCleary:
I didn't see that there were 4 other tests like this one. I'll add the 2 second range to those too

BrettCleary:
> LGTM but I am still not sure how a test like this, that is never exact and needs this type of check, really brings something to the client.

I think I was the one to write these originally and this was to fix issues with eta on the download toast. These tests ensure we don't get values like "00:00:1000" for 1000 seconds or negative values for downloaded > total size, etc. These were issues that made it through the QA process the first time around if I remember correctly. 

Note that the 2 second range still allows us to test and ensure the function is correct for these cases

PR #516, opened 2023-09-06
[CI] Add caching to CI workflows

flavioislima:
Tried using the actions/cachev3 first but I was seeing no improvement over the caching implementation of the setup/nodejs and it was also more complicated to setup. So I decided to use only the nodejs implementation.

---

### DIFF
--- a/src/backend/__tests__/utils.test.ts
+++ b/src/backend/__tests__/utils.test.ts
@@ -172,7 +172,7 @@ describe('backend/utils.ts', () => {
         getBytesFromMB(downloadSpeedMB),
         getBytesFromMB(totalSizeMB)
       )
-      expect(etaString).toEqual('00:00:09')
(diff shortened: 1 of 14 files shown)


src/backend/__tests__/utils.test.ts
BrettCleary, line 1: we'll need to revert this change after merging/rebasing main

BrettCleary (approved):
Cool! Looks like setup-node cache uses actions/cache under the hood with the yarn.lock hash as part of the key. That's a good strategy. We'll definitely get some CI improvements from this!

src/backend/__tests__/utils.test.ts
BrettCleary: nvm just handled it :ok_hand:

eliobricenov (approved):

PR #532, opened 2023-09-15
[Fix] Library Filtering

BrettCleary:
Fixes the alphabetical installed filter

Refactors library global context provider state into mobx

Adds unit tests for library state logic, which catches issues like the alphabetical/installed filter and prevents regression

---


src/frontend/state/libraryState.ts
```diff
@@ -0,0 +1,375 @@
...
+  libraryTopSection = ''
+  favouriteGames: GameCollection | undefined
+  hiddenGames: GameCollection | undefined
+  filterText = ''
+
+  // store
+  category: Category = 'all'
+
+  // filters
+  selectedFilter: DropdownItemType | undefined
```
eliobricenov, line 46: maybe we should define the filters as type? Something like:
```typescript
type Filter = 'asc' | 'desc'
const filterNames: Record<Filter, string> = {
    asc: 'Ascending'
    desc: 'Descending'
}
```

src/frontend/state/libraryState.ts
```diff
@@ -0,0 +1,375 @@
...
+  hyperPlayCategories,
+  sideloadedCategories
+} from 'frontend/helpers/library'
+
+class LibraryState {
+  epicLibrary: GameInfo[] = []
+  gogLibrary: GameInfo[] = []
+  sideloadedLibrary: GameInfo[] = []
+  hyperPlayLibrary: GameInfo[] = []
+  nonAvailbleGames: GameInfo[] = []
```
eliobricenov, line 32: typo here

src/frontend/state/libraryState.ts
```diff
@@ -0,0 +1,375 @@
...
+        : []
+      const HPLibrary = hyperPlayCategories.includes(this.category)
+        ? this.hyperPlayLibrary
+        : []
+
+      library = [...HPLibrary, ...sideloadedApps, ...epicLibrary, ...gogLibrary]
+
+      if (!this.showNonAvailable) {
+        // TODO: refactor to linear algo with non available game map
+        library = library.filter(
```
eliobricenov, line 237: copilot for the win:
```typescript
const nonAvailableAppNames = Object.fromEntries(this.nonAvailbleGames.map((game) => [game.app_name, true]))
library = library.filter((game) => !nonAvailableAppNames[game.app_name])
```

src/frontend/state/__tests__/libraryState.test.ts
eliobricenov, line 1: niiiiiceeeeeee 🔥

src/frontend/state/libraryState.ts
eliobricenov: We're already doing this for the category

eliobricenov (commented):
Awesome effort, Brett! 
I left some questions. Probably most of them are for code that was moved from place, but we might as well revisit them since we are refactoring

src/frontend/helpers/library.ts
```diff
@@ -99,7 +99,7 @@ async function install({
     installLanguage,
     runner,
     platformToInstall,
-    gameInfo,
+    gameInfo: JSON.parse(JSON.stringify(gameInfo)),
```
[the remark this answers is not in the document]
BrettCleary: yes. it gives `object could not be cloned` error if I don't clone it here. electron's ipcRenderer calls uses structured clone https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm which is giving errors here.

[this approach could give some data loss](https://stackoverflow.com/questions/122102/what-is-the-most-efficient-way-to-deep-clone-an-object-in-javascript) but I don't think we're using any of `Dates, functions, undefined, Infinity, RegExps, Maps, Sets, Blobs, FileLists, ImageDatas, sparse Arrays` on game info except undefined which is fine if we lose those props

src/frontend/state/libraryState.ts
BrettCleary: fixed now

src/frontend/state/libraryState.ts
BrettCleary: hmm. well the DropdownItemType is meant to be extensible to any dropdown item defined by the consuming code of the dropdown, so we can't type the id to be 
```
export type Filter = 'alphabeticalAscending' | 'alphabeticalDescending' | 'sortByInstalled'
```
in the ui library.

In the client I just added a `FilterItem` type that extends the dropdown item type and types the id prop to the filter values.

I think this addresses your concern. we only use the filter values in 3 or 4 lines in the code too so perfect typing isn't critical here imo

src/frontend/state/libraryState.ts
```diff
@@ -0,0 +1,375 @@
...
+      }
+    }
+
+    if (game.browserUrl && platformsToInclude.includes('browser')) return true
+
+    return false
+  }
+
+  filterByPlatform(library: GameInfo[], platformsToInclude: Platform[]) {
+    if (!library) {
```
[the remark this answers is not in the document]
BrettCleary: hmm true. I think I just copied the previous code. I'll remove this check

src/frontend/state/libraryState.ts
BrettCleary: nice. this is pretty clean. and linear :partying_face: just updated it

and we can confidently make changes like this now because we have tests :man_dancing:

src/frontend/state/libraryState.ts
```diff
@@ -0,0 +1,375 @@
...
+      library = library.filter(
+        (game) => !hiddenGamesAppNames?.includes(game?.app_name)
+      )
+    }
+
+    // sort
+    library = library.sort((a: { title: string }, b: { title: string }) => {
+      const gameA = a.title.toUpperCase().replace('THE ', '')
+      const gameB = b.title.toUpperCase().replace('THE ', '')
+      return this.sortAscending
```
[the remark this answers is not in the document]
BrettCleary: looks like prettier made it uglier by taking away my parenthesis and inserting all those new lines :cry: 

just updated

src/frontend/state/libraryState.ts
```diff
@@ -0,0 +1,375 @@
...
+  // top section
+  get showRecentGames() {
+    return this.libraryTopSection.startsWith('recently_played')
+  }
+}
+
+const libraryState = new LibraryState()
+export default libraryState
+
+export function resetLibraryState() {
```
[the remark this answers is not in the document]
BrettCleary: tru. just moved it

eliobricenov (approved):
LGTM thanks for addressing the comments

nyghtstalker:
@BrettCleary Unable to apply Favorite via the game card icon or menu option. Unable to apply hidden via the game card menu.

image image

nyghtstalker:
@BrettCleary Favorite and Hide are now working correctly but now I am unable to click into the library game card and "Details" on the menu does not work.

image

nyghtstalker (approved):
User is now able to click into the game page by clicking the game card.

PR #540, opened 2023-09-19
[UX] Add overlay toggle text

### LINKED ISSUE #538: Display instructions for toggling overlay on the overlay
Right now the overlay does not indicate to the player how to toggle it, we're only relying on a notification in the bottom right during the game launch.

If the player misses this, and they don't have the key combo memorized, they will either be unable to toggle the overlay, or stuck on it if it appears automatically.

BrettCleary:
Closes #538 

Windows
image

Linux
image

mac
image

---


src/frontend/OverlayManager/ExtensionManager/index.tsx
```diff
@@ -119,6 +119,7 @@ const BrowserExtensionManager = function ({
       {showOverlay ? (
         <>
           <div className={BrowserExtensionManagerStyles.bgFilter}></div>
+          <div className={`${BrowserExtensionManagerStyles.closeOverlayText} title`}>{t('hyperplayOverlay.closeOverlay', 'Press Alt or Option + X to close the overlay')}</div>
```
flavioislima, line 122: Could we split the message based on the platform?
So if it is macOS we show the OPT + X and for the rest we show ALT + X?

public/locales/en/gamepage.json
flavioislima, line 1: What happened with these json files? linting?
This might cause conflicts with the weblate branch.

src/frontend/OverlayManager/ExtensionManager/index.tsx
eliobricenov: Agree, I was confused the first time I used it.

BrettCleary: This could be confusing because you can use keyboards with an alt key on mac and then the keybind from the player's view is alt+X, not option + X, on mac.

public/locales/en/gamepage.json
BrettCleary: Ah I ran yarn i18n on windows and didn't have git setup to convert CRLF to LF line endings. Fixed now. back to LF line endings

eliobricenov (approved):
lgtm with the new text

nyghtstalker:
image

image

image

image

image

nyghtstalker (approved):
Text appears when the full-screen overlay is triggered.

src/frontend/OverlayManager/ExtensionManager/index.tsx
BrettCleary: updated the keybinds to be platform dependent. they use the same i18n strings but with markup variables

Issue #543, opened 2023-09-21
Bug: Wine/Crossover Version Selector shows on Windows


BrettCleary:
![image](https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/assets/2301666/9ef821c1-2575-43dd-9962-d8d60bf7c941)

nyghtstalker:
This is occurring on the current live version 0.6.0 on Windows.

PR #545, opened 2023-09-21
[Fix] Windows game settings showing wine/crossover

BrettCleary:
closes https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/issues/543

---

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -349,14 +349,10 @@ if (!gotTheLock) {
     ])
 
     // keyboards with alt and no option key can be used with mac so register both
-    const toggleOverlayHandler = () => {
-      toggleOverlay()
-      backendEvents.emit('OVERLAY_TOGGLED')
-    }
     const openOverlayAccelerator = 'Alt+X'
-    globalShortcut.register(openOverlayAccelerator, toggleOverlayHandler)
+    globalShortcut.register(openOverlayAccelerator, toggleOverlay)
     const openOverlayAcceleratorMac = 'Option+X'
-    globalShortcut.register(openOverlayAcceleratorMac, toggleOverlayHandler)
+    globalShortcut.register(openOverlayAcceleratorMac, toggleOverlay)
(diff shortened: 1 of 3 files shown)


eliobricenov (approved):

flavioislima (approved):
I am fine with this solution, although it will give conflicts with Heroic eventually.
A more cleaner approach would be to add a `isWindows` check on `WineVersionSelector` and return `null` if `true`. Since this is the only component that is being rendered on windows and should not.

eliobricenov:
> I am fine with this solution, although it will give conflicts with Heroic eventually. A more cleaner approach would be to add a `isWindows` check on `WineVersionSelector` and return `null` if `true`. Since this is the only component that is being rendered on windows and should not.

ohhhh I like this

BrettCleary:
> I am fine with this solution, although it will give conflicts with Heroic eventually. A more cleaner approach would be to add a `isWindows` check on `WineVersionSelector` and return `null` if `true`. Since this is the only component that is being rendered on windows and should not.

@flavioislima @eliobricenov 

We also need to hide the all the children of `Collapsible` here. It solely contains cross platform elements like `WinePrefix`, `CrossoverBottle`, etc.

GamesSettings is shown on both the settings pages and when you click into a game card's settings, so it is confusing to just add the `isWindows` check because `nativeGame` holds no meaning on the settings page and returns undefined. The dev will have to add logs and run to determine this or follow the window.api method and read through that. Also someone could change the return value for that window.api method for undefined parameters and it would break this conditional render in GamesSettings. This is too coupled and isn't very maintainable

The following is more verbose but prevents the above regression, makes the code more readable, and allows the dev to understand what is happening by purely reading the code instead of adding logs and running it
 ```
 const showCrossPlatformOptions = (isSettingsPage && !isWin) || (!isSettingsPage && !nativeGame)
 ```
 
 I'll add a comment above this line summarizing this

PR #557, opened 2023-09-27
[UX] Check access code on install modal

BrettCleary:
Closes https://github.com/HyperPlay-Gaming/product-management/issues/47

### Testing

here are some access codes to test the Kosium Beta channel with
H87YS0VDATNM
6915SC7P4UGY
0I0EJBVJI865
KFDC8T9RKT0H
WW6YGNBCUE9H
M2SSQMZFMO8E
J1W0871M1B9D
IAW4NTI2LBFA
OPBCBE50I1GB
2EJ2XQ0COTRN

### Screenshots

Initial modal render after selecting access code gated channel
image

Invalid State
image

Valid State
image

Non access code branch
image

---


src/frontend/screens/Library/components/InstallModal/index.tsx
```diff
@@ -153,6 +158,36 @@ export default React.memo(function InstallModal({
       setPlatformToInstall(availablePlatforms[0].value as InstallPlatform)
   }, [availablePlatforms])
 
+  const channelRequiresAccessCode =
+    !!selectedChannel?.license_config.access_codes
+
+  useEffect(() => {
```
eliobricenov, line 164: [I like to avoid useEffect's whenever it is possible](https://react.dev/learn/you-might-not-need-an-effect#sharing-logic-between-event-handlers). I see an opportunity here to use a form and display the error on submit. 

It'd be something like this:

```tsx
function handleSubmit(event: FormEvent) {
 if (needsAccessCode) {
   const valid = await window.api.validate()
    if (!valid) {
       setErrorText('invalid code')
       return
     }
  }
}
```
What do you think?

eliobricenov: We'd also show the install button always, it'd be always enabled and would trigger the form submit.

BrettCleary: hmm interesting. let me check with the designers

eliobricenov: Based on the conversation with the designer the only change remaining would be to make the install button always visible. I believe it should only be enabled when the code is valid.

BrettCleary:
updated states

first switch to access code gated branch 
image

incorrect code 
image

correct code
image

src/frontend/screens/Library/components/InstallModal/index.tsx
BrettCleary: sounds good. I did this and also added the red error stroke and the green success text

eliobricenov (approved):
Code LGTM!

I have two UX comments:

1 - I believe It'd be good to see why the button is disabled. That's why I think submit buttons should never be disabled and if there's an error, show it.

2 - Probably an edge case but if I remove the code it seems the validaiton is not triggered again

src/frontend/screens/Library/components/InstallModal/index.tsx
```diff
@@ -153,10 +159,56 @@ export default React.memo(function InstallModal({
...
+
   const showDownloadDialog = !isSideload && gameInfo
 
   const disabledPlatformSelection = Boolean(runner === 'sideload' && appName)
 
+  const enableCTAButton =
+    !channelRequiresAccessCode ||
+    (channelRequiresAccessCode && accessCodeVerified)
+
+  console.log('errorText ', errorText)
```
eliobricenov, line 210: oh a console.log slipped haha

BrettCleary:
> Code LGTM!
> 
> I have two UX comments:
> 
> 1 - I believe It'd be good to see why the button is disabled. That's why I think submit buttons should never be disabled and if there's an error, show it.
> 
> 2 - Probably an edge case but if I remove the code it seems the validation is not triggered again
> 
> Demo: https://www.loom.com/share/6c50a1bc2a1842afb85b2d65bc960399

I think for # 1, we can bring that up with the designers when they redesign the component. This PR already gets us 90% of the way there. Things like a popover on the button disable state can be handled on a separate branch

For 2, I just sent a fix. nice catch :+1:

src/frontend/screens/Library/components/InstallModal/index.tsx
BrettCleary: ah. just removed it :ok_hand:

src/frontend/components/UI/TextInputField/index.tsx
```diff
@@ -1,5 +1,5 @@
 import React, { ChangeEvent, FocusEvent, ReactNode, useContext } from 'react'
-import classnames from 'classnames'
+import { default as cn } from 'classnames'
```
flavioislima, line 2: A bit nitty-picky but normally we use this imported as `classNames` instead of `cn`.
Would be good to use the same one to make it easier to search in the future. Don't matter if `classNames`, `cn` or `cx`, but having the same its easier.

flavioislima (approved):
Looks good 👍🏽

src/frontend/components/UI/TextInputField/index.tsx
BrettCleary: sure thing. can't remember why I changed it

nyghtstalker (approved):
@BrettCleary I agree with Elio's comment about having something displayed as to why it isn't working. Helps the user understand what is going on instead of leaving it to them to guess. Maybe clicking when it is disabled and then it displays a text saying," Please enter a valid access code" in the same area with the other text? 

Other than that it looks good to me! Love the validation! =D

BrettCleary:
> @BrettCleary I agree with Elio's comment about having something displayed as to why it isn't working. Helps the user understand what is going on instead of leaving it to them to guess. Maybe clicking when it is disabled and then it displays a text saying," Please enter a valid access code" in the same area with the other text?
> 
> Other than that it looks good to me! Love the validation! =D

Sounds good. I will ping Billie about this next week and we can open a separate branch for it if needed

biliesilva:
The best approach here is to leave the button in an active state and show this message inside the access code input. If the user clicks on install without pasting/adding the code, it will not install because the code is required, but then we can show a warning message below the input component in red: “A code is required for this installation”, like this: 
(I tagged you guys on the Figma Game launcher UI file)

Modal - Installing a game


Modal - after pressing install without access code

PR #563, opened 2023-10-02
[TECH]: auth integration part 1

eliobricenov:
### Context

We need to create a modal that has webview linked to the dev portal signin page. 

The auth modal is only displayed once when the QA mode is activated. This is only for now and we will add more custom control when the achievement piece is added.

Dev portal PR: https://github.com/valist-io/hyperplay-dev/pull/125
Sub module PRs:
- https://github.com/HyperPlay-Gaming/hyperplay-extension-helper/pull/33#pullrequestreview-1709315466
- https://github.com/HyperPlay-Gaming/hyperplay-proxy-server/pull/26
- https://github.com/HyperPlay-Gaming/hyperplay-overlay/pull/5

**Steps for manual testing available in Qase: https://app.qase.io/project/HC?case=144&suite=31**
**Simple flow of the UX:  https://app.skiff.com/file/66246748-6d6f-43d5-9b50-f61c6e3cd359**
**Figjam of the UX: https://www.figma.com/file/Bp8NMouDfb8N08Wc9lEfpl/auth-UX-flow?type=whiteboard&t=iKQoIfHUeg3kVqKu-0**

# Testing

**Be sure to test with the packaged app so that the email confirmation deeplinks will work**


src/frontend/state/authModalState.ts
```diff
@@ -0,0 +1,34 @@
+import { makeAutoObservable } from 'mobx'
+
+class AuthModalState {
+  private qaModeActive = false
+  private open = true
```
BrettCleary, line 5: nitpick but it might be good to use private fields instead of private keyword if you want to be sure this is not modified from outside the class at runtime.

This is what I've started to do with my new mobx classes

https://stackoverflow.com/questions/59641564/what-are-the-differences-between-the-private-keyword-and-private-fields-in-types
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

src/frontend/components/UI/AuthModal/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+  return error.includes(METAMASK_ALREADY_PROVIDED_ERROR_CODE.toString())
+}
+
+const AuthModal = () => {
+  const webviewRef = useRef<WebviewTag>(null)
+
+  const handleAccountNotConnected = async () => {
+    const currentProvider = await window.api.getCurrentWeb3Provider()
+
+    if (!currentProvider) {
```
BrettCleary, line 27: there is an `UNCONNECTED` state in the `PROVIDERS` enum. we should open onboarding for that value too

src/frontend/components/UI/AuthModal/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+      switch (event.channel) {
+        case 'closeAuthModal':
+          authModalState.closeModal()
+          break
+        case 'auth:accountConnected':
+          alert('account connected')
+          authModalState.closeModal()
+          break
+        case 'auth:accountNotConnected':
+          // TODO: try to resume flow after connecting account
```
BrettCleary, line 60: did you want to handle that on this branch?

src/frontend/components/UI/AuthModal/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+const isTooManyRequestsError = (error: string) => {
+  return error.includes(METAMASK_ALREADY_PROVIDED_ERROR_CODE.toString())
+}
+
+const AuthModal = () => {
+  const webviewRef = useRef<WebviewTag>(null)
+
+  const handleAccountNotConnected = async () => {
+    const currentProvider = await window.api.getCurrentWeb3Provider()
```
BrettCleary, line 25: you actually want to use `window.api.getConnectedProvider()` here. `getConnectedProvider()` returns the currently connected provider

`getCurrentWeb3Provider()` returns the one the user last connected with so we can auto-connect with that on subsequent launches

src/frontend/components/UI/AuthModal/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+    if (!currentProvider) {
+      onboardingStore.openOnboarding()
+      return
+    }
+
+    // try to trigger metamask popup to connect account
+    try {
+      const provider = new ethers.BrowserProvider(window.ethereum)
+      await provider.getSigner()
+    } catch (e) {
```
BrettCleary, line 36: I think the flow you want here is 

if there is no connected provider or current provider, then call `onboardingStore.openOnboarding()` 

if there is no connected provider but there is a current provider, then call
```
const currentWeb3Provider = await window.api.getCurrentWeb3Provider()
onboardingStore.bootstrapOnboarding(context, currentWeb3Provider)
``` 

if there is a connected provider (there will always be a current provider too in this instance), then you're good and 
```
window.ethereum.request({ method: 'eth_requestAccounts' })
```

should return the accounts

src/frontend/components/UI/AuthModal/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+const METAMASK_ALREADY_PROVIDED_ERROR_CODE = -32002
+
+const isTooManyRequestsError = (error: string) => {
+  return error.includes(METAMASK_ALREADY_PROVIDED_ERROR_CODE.toString())
+}
+
+const AuthModal = () => {
+  const webviewRef = useRef<WebviewTag>(null)
+
+  const handleAccountNotConnected = async () => {
```
BrettCleary, line 24: after calling the flow I detailed below, you could return a promise and when `window.api.handleAccountsChanged(cb)` calls cb, you can be sure it's connected and resolve. then you can go on and resume the auth modal connection stuff

src/frontend/components/UI/AuthModal/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+const url =
+  'https://hyperplay-dev-git-feature-unified-auth-ui-valist.vercel.app/signin'
+
+const METAMASK_ALREADY_PROVIDED_ERROR_CODE = -32002
+
+const isTooManyRequestsError = (error: string) => {
+  return error.includes(METAMASK_ALREADY_PROVIDED_ERROR_CODE.toString())
+}
+
+const AuthModal = () => {
```
BrettCleary, line 21: thoughts about moving some of this component state into the `authModalState` mobx class? for achievements, we may want to observe an `isHyperPlayAccountSignedIn` from `authModalState` to update ui elements. Or a variable indicated which stores are connected to the account

src/frontend/components/UI/AuthModal/index.tsx
BrettCleary: actually you could probably use `WalletStore.address` and let mobx trigger a re-render. that's a lot cleaner

BrettCleary (changes requested):
nice work. welcome to the world of complex provider boolean logic 🙃  haha

BrettCleary:
I'm wondering what we want the behavior to be when the user hasn't gone through MM onboarding yet. for instance, this is what I got when clicking `MetaMask` 

so I'm forced to close out of the onboarding modal. then I'll have to go back and try to sign in again after I onboard, which isn't the best UX. It is probably good enough for MVP though, just making a note of it here

image

src/frontend/state/authModalState.ts
BrettCleary, line 5: we should use `window.api.handleQaModeActivated` to set open back to true. otherwise, I have to close and reopen HyperPlay to trigger the auth modal again

BrettCleary:
After signing the SIWE request with mm extension, the modal is still open. is there supposed to be a reaction somewhere to let me know it was successful?

src/frontend/components/UI/AuthModal/index.tsx
BrettCleary: When clicking mm mobile or wallet connect, mm extension is still requesting to sign the SIWE message.

We should update the above flow to compare the connected/current provider to the requested provider from the auth modal

BrettCleary:
when clicking the email sign up button, I receive no email and there is no feedback from the auth modal. is email hooked up?

eliobricenov:
> After signing the SIWE request with mm extension, the modal is still open. is there supposed to be a reaction somewhere to let me know it was successful?

It seems there is something wrong with the Auth API in Vercel. Let me check in with Alec and get back to you.

eliobricenov:
> when clicking the email sign up button, I receive no email and there is no feedback from the auth modal. is email hooked up?

No, it is not yet. We'll hook that part with the OAuth part.

src/frontend/components/UI/AuthModal/index.tsx
eliobricenov: I was planning to do it in a different branch but with your suggestions, I may have an idea of how to do it.

src/frontend/state/authModalState.ts
eliobricenov: Oh it seems there is a limitation on MobX https://mobx.js.org/observable-state.html#limitations 

> EcmaScript private fields (#field) are not supported. When using TypeScript, it is recommended to use the private modifier instead.

BrettCleary: ah gotcha. I guess that's only for observables but that could get confusing quick. cool. let's stick with the private keyword then

src/frontend/state/ExtensionHandlerState.ts
```diff
@@ -0,0 +1,165 @@
+class ExtensionHandlerState {
```
eliobricenov, line 1: kudos to @BrettCleary for this refactor!

BrettCleary:
Could we automatically request the SIWE signature when the app refreshes? 
this PR https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/pull/607 reopens the auth modal, which takes out one step, but the user still has to click mm extension again

I am thinking we can use the same `openAuthModalIfAppReloads` flag and then have a window.api method that the auth modal calls on initial launch.

what do you think @eliobricenov ?

eliobricenov:
> Could we automatically request the SIWE signature when the app refreshes? this PR #607 reopens the auth modal, which takes out one step, but the user still has to click mm extension again
> 
> I am thinking we can use the same `openAuthModalIfAppReloads` flag and then have a window.api method that the auth modal calls on initial launch.
> 
> what do you think @eliobricenov ?

Do you mean after the user onboarding with MM? We'd still the user to connect the wallet for the first time I think.

src/frontend/components/UI/AuthModal/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+    const handleDomReady = () => {
+      webview.addEventListener('ipc-message', handleIpcMessage)
+    }
+
+    webview.addEventListener('dom-ready', handleDomReady)
+
+    const qaModeListenerCleanup = window.api.handleQaModeActivated(() => {
+      authState.activateQaMode()
+      authState.openSignInModal()
```
red-game-dev, line 86: This we should care, if it is something only for development, staging, PTR versions, we should make a check for it depending on env for this. If this gets exposed to prod, it can be triggered.

eliobricenov: good call let's add another flag with a constant

BrettCleary: @eliobricenov we feature flag amazon game store in `src/frontend/constants.ts`
with 
`export const ENABLE_AMAZON_STORE = false`

we could do that or use the new experimental flag that flavio added in the global context provider

BrettCleary (approved):
double checked and all manual tests are passing. code looks good! 🥳

src/common/constants.ts
```diff
@@ -7,3 +7,6 @@ export const devSentryDsn =
 export const METAMASK_SNAPS_URL = 'https://snaps.metamask.io/'
 export const METAMASK_SNAPS_EXTERNAL_URL = 'https://metamask.io/snaps/'
 export const METAMASK_PORTFOLIO_URL = 'https://portfolio.metamask.io/'
+
+export const DEV_PORTAL_URL =
+  'https://hyperplay-dev-git-feature-unified-auth-ui-valist.vercel.app/'
```
red-game-dev, line 12: I think this needs to be changed?

src/frontend/components/UI/AuthModal/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+import authState from '../../../state/authState'
+import { ethers } from 'ethers'
+import extensionState from '../../../state/ExtensionState'
+import onboardingState from '../../../store/OnboardingStore'
+import walletState from '../../../state/WalletState'
+import { DEV_PORTAL_URL } from 'common/constants'
+
+const url = `${DEV_PORTAL_URL}/signin`
+
+const METAMASK_ALREADY_PROVIDED_ERROR_CODE = -32002
```
red-game-dev, line 15: For later, we can always have a file where that includes Metamask error codes at

red-game-dev (approved):
As long we make sure that `DEV_PORTAL_URL` is proper, and the stringifying of session logout its fine, we can merge 🙌

src/common/constants.ts
BrettCleary: yes after we merge the dev portal pr

src/frontend/components/UI/AuthModal/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+    <ModalAnimation
+      isOpen={authState.isSignInModalOpen}
+      onClose={() => authState.closeSignInModal()}
+    >
+      <webview
+        ref={webviewRef}
+        src={url}
+        className={styles.webview}
+        partition="persist:auth"
+        allowpopups={'true' as unknown as boolean | undefined}
```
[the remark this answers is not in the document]
BrettCleary: it is an electron bug. took me a while to find it a while back on a previous PR

here it is with just `allowpopups`
image

here it is with `allowpopups={'true' as unknown as boolean | undefined}`
image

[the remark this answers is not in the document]
red-game-dev: Interesting, yup no worries that 🙌 Cheers for the detailed reply 🍻

nyghtstalker:
1. I am still working on exactly how to trigger this but the wallet unlock modal will open, not go away, and can't be closed.

image

image

2. When you click the MM icon to open the wallet unlock it appears on top of the other modal.

nyghtstalker:
It is showing as connected but still asking me to sign-in. It is also not showing under Remote Link.

image
image

eliobricenov:
> It is showing as connected but still asking me to sign-in.
> 
> image

We'll probably change this but for MVP I think this is fine

nyghtstalker:
The modals seem inconsistent/finicky and need some polish. The email modal flashes prior. 

https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/assets/82125535/6291e297-2773-42f3-af6a-0b9ca84c76c5

https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/assets/82125535/7419e37b-53d6-4b94-a431-38ace8377bc6

nyghtstalker:
Wallet connect buttons do not work after selecting MM Mobile option in the onboarding wallet modal. @BrettCleary @eliobricenov 

1. Make sure no wallet is connected
2. Wallet menu -> Manage Accounts (auth modal appears)
3. Select MM Mobile (Onboarding wallet modal appears)
4. Select MM Mobile so the QA code appears
5. Close wallet modal and go back to auth modal
6. Observe that the first three wallet connects do not work when clicked

nyghtstalker:
This may be the same issue as the previous one I mentioned but I found a similar instance but this time with the wallet attached and unlocked.

**1. Make sure the wallet is ATTACHED and UNLOCKED**
2. Open Account Manage Accounts
3. Select any of the first three wallet options -> this will open the wallet sign request **(DO NOT SIGN!)**
4. With the sign request still open select a different wallet option
5. Observe wallet options will now be inaccessible

https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/assets/82125535/b602edfb-44ed-464f-aebe-b2b894b75008

eliobricenov:
> This may be the same issue as the previous one I mentioned but I found a similar instance but this time with the wallet attached and unlocked.
> 
> **1. Make sure the wallet is ATTACHED and UNLOCKED** 2. Open Account Manage Accounts 3. Select any of the first three wallet options -> this will open the wallet sign request **(DO NOT SIGN!)** 4. With the sign request still open select a different wallet option 5. Observe wallet options will now be inaccessible
> 
>  2023-12-06_10-19-05.mp4

In this case, the signature request are triggered, and the notification count increased. The metamask popup does not show because it was already dismissed by the user. This is the expected behavior.

nyghtstalker (approved):
Validated that issues have been fixed. SHIP IT!

PR #584, opened 2023-10-13
[TECH] Add Win10 runner to e2e tests

awantoch:
We should refactor this to use matrix runners, but for now this adds Win10 test runners to e2e dev and packaged


flavioislima (approved):
LGTM
Not sure why we took so long to review this one. 😅

BrettCleary:
Do we need this? I don't think it provides much benefit and may be starting to ask too much of Github's free services and their acceptable use policies 
https://docs.github.com/en/site-policy/acceptable-use-policies/github-acceptable-use-policies#9-excessive-bandwidth-use 
https://docs.github.com/en/site-policy/github-terms/github-terms-for-additional-products-and-features#actions

BrettCleary:
closing this as I think we should just stick with the latest windows runner to test with. The download issue we were trying to test with win10 vs win latest turned out to not be a win10 issue

Issue #590, opened 2023-10-17
[CI/CD] Update Windows Release


BrettCleary:
A couple small fixes are needed here to completely automate the windows release

1. The draft release name needs to be changed on the windows release workflow to `0.7.0` to match the rest of the artifacts. Right now it is `Release 0.7.0` and then we have to copy the artifacts over
2. The windows release is named `HyperPlay-Setup.exe` instead of `HyperPlay-0.8.0-Setup-x64.exe`
3. The latest.yml needs the release name updated also to `HyperPlay-0.8.0-Setup-x64.exe`
4. The latest.yml needs the sha512 hash to be all lower case. It is currently all upper case which would break auto update on windows

flavioislima:
1. Now sure how to do that, using the powershell extension from Signpath it could not recognize something like `HyperPlay*.exe` like before when uploading the artifact.
4. SHASUM is not case sensitive, if auto-update on windows is not working then its another reason. But I tried on win11 and worked fine.

PR #595, opened 2023-10-18
[CI] Fix Windows and Snap workflows

flavioislima:
- Do not create a new release on Windows workflow, use the current draft release instead.
- Update the Linux Snap package only after the release was published.

---


.github/workflows/release_windows.yml
```diff
@@ -88,24 +88,17 @@ jobs:
...
+      - name: Upload asset
         id: upload-release-asset
         uses: actions/upload-release-asset@v1
         env:
           GITHUB_TOKEN: ${{ secrets.WORKFLOW_TOKEN }}
         with:
-          upload_url: ${{ steps.create_release.outputs.upload_url }}
+          upload_url: ${{ steps.get_release.outputs.upload_url }}
           asset_path: signed/HyperPlay-Setup.exe
           asset_name: HyperPlay-Setup.exe
```
BrettCleary, line 103: can we change these names on this branch? or is it out of scope?

.github/workflows/release_windows.yml
```diff
@@ -88,24 +88,17 @@ jobs:
           $latestYml.files[0].size = $size
```
BrettCleary, line 88: the artifact path would need to be updated here as well if we change the exe name

BrettCleary: can we lowercase the sha512 here? it is part of the AC here https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/issues/590

BrettCleary: i am referring to lines 82-88 btw. github won't let me highlight them so this is as close as i can get

.github/workflows/release_windows.yml
```diff
@@ -88,24 +88,17 @@ jobs:
...
-        env:
-          GITHUB_TOKEN: ${{ secrets.WORKFLOW_TOKEN }}
-        with:
-          tag_name: ${{ github.ref }}
-          release_name: Release ${{ github.ref }}
-          draft: true
-          prerelease: false
+      - name: Get release
+        id: get_release
+        uses: bruceadams/get-release@v1.3.2
```
BrettCleary, line 93: it seems like this GHA is meant to be used with the 

```
on:
  release:
    types: [created]
```

trigger. will it work if we're using the tag push trigger? Also will it work if this is the first workflow to complete (i.e. that there is no draft release created yet) so it has to create it?

https://github.com/bruceadams/get-release

BrettCleary: I think the previous actions/create-release would work if we removed the `Release ` prefix on release_name right? did you push a tag to this branch to try and test that?

BrettCleary (changes requested):

.github/workflows/release_windows.yml
flavioislima: I remember now why I needed to rename it like this.,
I was unable to make powershell to understand something like `HyperPlay*.exe` when uploading the file.
Tried in several ways and did not work.

flavioislima: Here on this part:
https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/blob/d0d1849d5698928c5022c87750bb3e49521dde49/.github/workflows/release_windows.yml#L82

.github/workflows/release_windows.yml
flavioislima: SHA sums are not case-sensitive. I had this question as well:
https://superuser.com/questions/1712250/do-lower-and-capital-letters-matter-for-sha-checksum

.github/workflows/release_windows.yml
flavioislima: If this is the first workflow to complete it won't work but it is basically impossible for it to be the first release. If it happens it's because the Linux one failed.
It takes around 20 minutes to complete while the Linux one is the first to create the release since it has no signing or anything, in 5 minutes we already have the draft release ready.

flavioislima: > I think the previous actions/create-release would work if we removed the `Release ` prefix on release_name right? did you push a tag to this branch to try and test that?

no, didn't know about that. We can use it but I don't think will make any difference in the end.
Looking at the code I don't see how it is triggered only on `created`, on their github they have only an example.

flavioislima:
@BrettCleary not sure how to achieve the changes on the artifact name, feel free to give it a try.
This is their documentation if you want to give it a try:
https://about.signpath.io/documentation/powershell/Submit-SigningRequest

flavioislima:
pushing a tag now to test if the changes will work.
I am not fluent in Powershell so I used Copilot to try to fix the artifact name.

.github/workflows/release_windows.yml
flavioislima: Tested and it creates a new release anyway. Looking for alternatives now.

flavioislima: Tried using some logic to check for the release and it was failing anyway. 
I think we should just use the Get Release action here anyway, if you check the times, at the point that the action starts building the windows exe, all the linux files were built and uploaded to the release, macOS as well. Windows is always the last one on this so should be safe enough.
There is also the problem of our quota on Signpath that is reaching the limit during these tests.

.github/workflows/release_windows.yml
BrettCleary: yes but it fails the check client side and refuses to auto update on windows if the cases are not identical. I believe this check is occurring in the `electron-updater` `autoUpdater`. Right now I have to manually lower case the hash before each release

.github/workflows/release_windows.yml
BrettCleary: can you use something like `HyperPlay-${{ github.ref }}-Setup-x64.exe`?

BrettCleary: or set the `${{ github.ref }}` as an env variable for that step and use that env variable in the exe string name

.github/workflows/release_windows.yml
BrettCleary: okay that works

.github/workflows/release_windows.yml
flavioislima: I fixed the naming already by getting the result of the listing. Because then we won't risk breaking the flow in case we update the artifact name in the future.

.github/workflows/release_windows.yml
flavioislima: thats really weird though. I will investigate this further.

flavioislima: ok, should be fixed now. The string will be lower case after being generated

BrettCleary (approved):
LGTM

eliobricenov (approved):
lgtm (had to do research + chatgpt to understand hehe)

PR #603, opened 2023-10-23
Feature flag Amazon Game Store

BrettCleary:
Add feature flag for AGS. Disable it.

---

### DIFF
--- a/src/frontend/App.tsx
+++ b/src/frontend/App.tsx
@@ -29,6 +29,7 @@ import DownloadToastManager from './components/UI/DownloadToastManager'
 import TopNavBar from './components/UI/TopNavBar'
 import StoreNavHandler from './StoreNavHandler'
 import QaAuthHandler from './QaAuthHandler'
+import { ENABLE_AMAZON_STORE } from './constants'
 
 function App() {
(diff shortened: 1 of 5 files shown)


nyghtstalker:
- Amazon has been removed from the Store, Library, and Manage Accounts pages.

image

image

image

nyghtstalker (approved):
Good job! Amazon has VANISHED!

eliobricenov (approved):
lgtm!

I'm curious why you decided to replace `&&` with ternaries? 👀

BrettCleary:
> lgtm!
> 
> I'm curious why you decided to replace `&&` with ternaries? 👀

@eliobricenov 
ah this is a good article but they paywalled it https://medium.com/geekculture/stop-using-for-conditional-rendering-in-react-a0f7b96200f8

basically if you render with 
```
return {false && <div>something</div>}
```

it'll render a 0 in the DOM because false is rendered and converted into a number

if you use a ternary it will just return null and nothing will be rendered

maybe we haven't been seeing theses 0's in this case because mantine filters them out, but we should use ternary anyways. it is less error prone. I have seen these random 0's from `&&` before and they can be pretty hard to track down

Issue #455, opened 2023-08-10
Install modal does not display extracting


nyghtstalker:
### Describe the bug

While a game is extracting the install modal still displays downloading. This should change to extracting.

Needs to display on the game card and downloads page too.

### Add logs

N/A

### Steps to reproduce

1. Install any game
2. Observe install modal shows downloading
3. Wait for the game to extract
4. Observe install modal still shows downloading

### Expected behavior

Install modal displays extracting

### Screenshots

![image](https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/assets/82125535/899c940b-95e8-4dca-9b85-e099345dcbd9)

![image](https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/assets/82125535/8d2bbb98-a47d-4d57-91cc-cbd0bdb9d9f3)

![image](https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/assets/82125535/5a7da6ff-09ce-45fc-b984-644b547e10d1)

### hyperplay Version

Latest Stable

### System Information

- OS: all
- hyperplay Version 0.4.2

flavioislima:
@red-game-dev maybe instead of 'extracting' we could just show 'installing' or 'verifying' like Steam does. It is more user friendly than extracting imo.

red-game-dev:
@flavioislima, we can discuss this further with @BrettCleary. Personally either of the two are fine, as letting know the user is being extracted at the moment, but perhaps for large downloaded game usually people does not really check the screen all the time, so less statuses would be better as we already have got some complexities in middle both from BE & FE 🙌 Let me know if we should discuss all this through call 🍻

Issue #614, opened 2023-10-30
Disk space check does not account for extraction space


nyghtstalker:
### Describe the bug

The disk space check only accounts for the install amount and not the extraction amount too. In the case of Phantom Galaxies it 23GB. When you add another 23GB for the extraction you will need 46GB of space free on the drive. The check should account for double the install amount to account for the needed space for adding the extraction.

### Add logs

N/A

### Steps to reproduce

1. Install any game
2. Observe the space checker on accounts for the install amount 

### Expected behavior

Disk check should account for the extraction amount too

### Screenshots

![image](https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/assets/82125535/39d764a3-8501-40f5-a071-3d270087540b)

### hyperplay Version

Latest Stable

### System Information

- Windows 10
- hyperplay Version 0.8.1

nyghtstalker:
Fixed in Main and waiting for next release.

nyghtstalker:
No longer occurs.

PR #536, opened 2023-09-18
[Fix] Preserve symlinks during extraction

BrettCleary:
Recreates symlinks after a zip file is extracted

### Testing
Test with synergy of serra native mac build latest main release
- `Synergy of Serra Genesis.app/Contents/Frameworks/Electron Framework.framework/*` are all simlinks. You can verify with ls -l which will return `lrwxrwxrwx` in the first column. `l` is for "link"

---

Use the following Checklist if you have changed something on the Backend or Frontend:

- [ ] Tested the feature and it's working on a current and clean install.
- [ ] Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)
- [ ] Created / Updated Tests (If necessary)
- [ ] Created / Updated documentation (If necessary)

PR #618, opened 2023-11-01
[UX] Improve Download Toast, Disk Space Calculation for Uncompressed & Relaunch Download Toast Issue

red-game-dev:
Closes: #455 

What is done: Having an extraction progress for Toast Download, Game Page, Game Card, along some issues should be sorted which you can check on `Issue Addressed`

Note: This pr does not include E2E testing

**What is the issue?**
Feel free to check & write on the thread on slack for this: https://hyperplay-workspace.slack.com/archives/C05R8CK6UCA/p1699036063002419

---

## 🐛 Issue Addressed

**Linked Issue:** [Install modal does not display extracting](https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/issues/455) - PR Main issue

---

## ✅  Other Fixed Issues

-  Progress on Download does not start from 0, generally was started either from previous queued progress if it was canceled earlier
- Disk Space is another issue mentioned by @nyghtstalker, and fixed with suggest of @flavioislima. Currently, we are utilizing a basic function to estimate disk space. Typically, the estimated size is doubled and then increased by 0.05-0.01%, depending on whether it is being compressed on a Mac (higher percentage) or on Windows and other operating systems. **Linked Issue:** [Disk space check does not account for extraction space](https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/issues/614)
- Upon relaunch the download modal appears again even though there is nothing downloading. This should be fixed, although @nyghtstalker, please make sure to check it out as soon this PR read for review
- Clicking Play button right after starting an uninstall disables the Install/Play button in the library game. **Linked Issue:** [Clicking Play button right after starting an uninstall disables the Install/Play button in the library game page](https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/issues/619)
- Installing after uninstalling, may have the chance to unable to download again, as nothing happens (may have chances to break on `rmAsync` method, which breaks the backend and disallow to reinstall again )

---

### ⚠️ Review Tests

The tests cases for this PR can be found at
https://app.qase.io/project/HC?previewMode=side&suite=43


red-game-dev:
A note: Will require to apply the changes from https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/pull/536/files , as soon is merged

flavioislima (commented):
So far the code looks good. Left a few comments. I believe that we need a full regression test on all stores and with games with DLCs as well to make sure that nothing is broken.

src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -551,83 +581,64 @@ export async function install(
...
-      return { status: 'error', error: `${error}` }
+    if (!platformInfo.executable) {
+      return {
+        status: 'error',
+        error: 'Executable not found during install in HyperPlay game manager'
+      }
     }
-    return { status: 'done' }
+
+    return await extract(appName, {
```
flavioislima, line 631: Not sure if we need `await` here since no other operation will run afterward

src/frontend/hooks/hasStatus.ts
```diff
@@ -9,7 +9,7 @@ import libraryState from 'frontend/state/libraryState'
 // the consuming code needs to be wrapped in observer when using this hook
 export const hasStatus = (
   appName: string,
-  gameInfo: GameInfo,
+  gameInfo?: GameInfo,
```
flavioislima, line 12: Why the change here?
Will it work if we do not pass gameinfo here?

e2e/__specs__/hpStoreApi.spec.ts
```diff
@@ -94,6 +93,43 @@ test.describe('hp store api tests', function () {
     }
   }
 
+  // const cancelExtraction = async (appName: string) => {
```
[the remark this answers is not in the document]
red-game-dev: This is commented temporarily, as trying to fix the issue of E2E, that the progress loops without status. If uncommented the code would run the pipeline until timeout

src/frontend/components/UI/DownloadToastManager/index.tsx
```diff
@@ -29,44 +26,22 @@ export default function DownloadToastManager() {
...
-
-  const [dmState, setDMState] = useState<DownloadManagerState>('idle')
-
-  useEffect(() => {
-    window.api.getDMQueueInformation().then(({ state }: DMQueue) => {
-      setDMState(state)
-    })
+  const appName = currentElement?.params.gameInfo.app_name
+    ? currentElement?.params.gameInfo.app_name
+    : ''
```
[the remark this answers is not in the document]
red-game-dev: Cheers for suggestion, will need to apply this manually though 🍻 

image

src/frontend/hooks/hasStatus.ts
red-game-dev: Initially, it was from @BrettCleary  as the following, therefore just changed. I believe this I changed as soon I went on this PR 

```
export const hasStatus = (
  appName: string,
  gameInfo: GameInfo | undefined,
```

src/backend/storeManagers/hyperplay/games.ts
red-game-dev: True, although eslint will complain than, generally it should return extract(...) and since the function is async, than it will await for it still as it will return a promise
image

flavioislima: Our eslint config is going too far 😂

e2e/__specs__/hpStoreApi.spec.ts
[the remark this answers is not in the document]
eliobricenov: As a side note you you can do `test.skip` if you don't want to comment out the code

red-game-dev:
> So far the code looks good. Left a few comments. I believe that we need a full regression test on all stores and with games with DLCs as well to make sure that nothing is broken.

Certainly, a full regression is required for this

e2e/__specs__/hpStoreApi.spec.ts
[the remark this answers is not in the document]
red-game-dev: Oh ye, could use that for now, just went with the classic way of commenting 😆

src/backend/storeManagers/hyperplay/games.ts
red-game-dev: Sure is 😆 I was thinking when I get that warning `why, but why` 🤣

BrettCleary:
@red-game-dev let's move these into qase 
image

BrettCleary:
Looks like we have some build, test, codecheck errors. I think you need to merge the overlay changes from main @red-game-dev

red-game-dev:
> @red-game-dev let's move these into qase image

I saw that can create suites & cases there, just before starting creating, would like to confirm somethings

Shall we have `Install Model` Suite under project root, and others separately like `Game Card > Download Bar`, `Download Toast`  or we would have this all under Store Page as Installation with different cases to test?

src/frontend/hooks/useEstimatedUncompressedSize.ts
```diff
@@ -0,0 +1,11 @@
...
+  platform: string,
+  compressedSize: number
+) => {
+  const baseEstimate = compressedSize * 2
+  const gapPercentage = platform === 'osx' ? 0.05 : 0.1
+
+  const gap = baseEstimate * gapPercentage
+
+  return baseEstimate + gap
+}
```
BrettCleary, line 11: what is the behavior if this over or underestimates the uncompressed size?

src/frontend/hooks/useGetDmState.ts
BrettCleary, line 1: can we use `src/frontend/state/DMQueueState.ts` instead of this?

src/frontend/hooks/useGetDownloadStatusText.ts
BrettCleary, line 1: could this go into a mobx class like DMQueueState also?

src/frontend/screens/Game/GamePage/index.tsx
```diff
@@ -140,6 +141,11 @@ export default observer(function GamePage(): JSX.Element | null {
 
   const storage: Storage = window.localStorage
 
+  const uncompressedSize = useEstimatedUncompressedSize(
+    platform,
+    gameInstallInfo?.manifest?.disk_size || 0
+  )
```
BrettCleary, line 147: maybe we should only guess the uncompressed size if `installSize !== downloadSize` i.e. if we actually have the value

src/frontend/hooks/hasStatus.ts
BrettCleary: @red-game-dev @flavioislima looks like this propagates from src/frontend/component/UI/DownloadToastManager where the currentElement could be undefined. It is set to undefined after a timeout when it finishes installing to reset the toast. When currentElement or gameInfo are undefined, `<></>` is returned so no worries about showing the wrong state on the download toast

iirc we can't return earlier than `hasStatus` otherwise we change the number of hooks rendered on subsequent re-renders when a download starts

Honestly we should just move all this state and logic into a mobx class and add tests for it. Then we wouldn't have to deal with these annoying react limitations. This code was written before we committed to mobx.

BrettCleary (changes requested):
good stuff! love the tests 💜

BrettCleary:
> > @red-game-dev let's move these into qase image
> 
> I saw that can create suites & cases there, just before starting creating, would like to confirm somethings
> 
> Shall we have `Install Model` Suite under project root, and others separately like `Game Card > Download Bar`, `Download Toast` or we would have this all under Store Page as Installation with different cases to test?

What do you think about a test suite structure like this?

Game Installation (Test Suite)
 - Download Toast (Test Suite)
   - Clicking pause on the download toast pauses the game download (test case)
   - Clicking cancel on the download toast cancels the game download (test case)
 - Download Manager (Test Suite)
   - ...
 -  Game Card (Test Suite)
   - ...
 - Game Page (Test Suite)
   -

src/frontend/hooks/useEstimatedUncompressedSize.ts
BrettCleary, line 1: if this is a placeholder, let's add a `//TODO:...` comment and create an issue for it so we don't forget to update it when our patching service is live

src/frontend/hooks/useEstimatedUncompressedSize.ts
BrettCleary: is it used to calculate the % extraction? as in `sizeOfExtractedFiles / estimatedUncompressedSize * 100 = percentExtracted`? or is there another way we're getting the denominator?

src/frontend/hooks/useGetDownloadStatusText.ts
BrettCleary: gotcha. no problem. this works then

src/frontend/hooks/useEstimatedUncompressedSize.ts
red-game-dev: Does not calculate the extraction %, thats only at the BE which we receive through `window.api.onProgressUpdate`, just the installation model or game page `install size` and does not affect any other logic around: 

**Below is where is being implemented**

Installation Model
image

Game Page
image

red-game-dev: Also, I didn't include these tests cases in https://app.qase.io/project/HC?previewMode=side&suite=43, but I can include them. btw, this is the issue linked to this change: https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/issues/614

src/backend/__tests__/utils.test.ts
```diff
@@ -257,59 +247,4 @@ describe('backend/utils.ts', () => {
...
-      console.log('extracting test.zip')
-    })
-
-    test('extract a test zip with non ascii characters', async () => {
-      const renamedZipFilePath = path.resolve(
-        './src/backend/__mocks__/谷���新道ひばりヶ�.zip'
-      )
-      renameSync(testCopyZipPath, renamedZipFilePath)
-      testCopyZipPath = renamedZipFilePath
-    })
```
BrettCleary, line 307: This is an important test case as we have had games with non ascii characters fail to extract before. I do not see it in `ExtractZipService.test.ts`. Can you add it?

src/backend/__tests__/utils.test.ts
```diff
@@ -257,59 +247,4 @@ describe('backend/utils.ts', () => {
...
-
-    it('should throw an error if the zip file does not exist', async () => {
-      await expect(
-        extractZip('nonexistent.zip', destFilePath)
-      ).rejects.toThrow()
-    })
-  })
```
BrettCleary, line 314: I also do not see this test case represented in your `ExtractZipService.test.ts` test cases. Can you add it?

src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -472,13 +494,35 @@ async function resumeIfPaused(appName: string): Promise<boolean> {
...
+}
+
 export async function install(
   appName: string,
   { path: dirpath, platformToInstall, channelName, accessCode }: InstallArgs
 ): Promise<InstallResult> {
-  if (await resumeIfPaused(appName)) {
-    return { status: 'done' }
-  }
+  const wasPaused = await resumeIfPaused(appName)
```
BrettCleary, line 524: Could we not check if an extraction was in progress in `resumeIfPaused` and handle the resuming of the extraction there? I think it is confusing to have a wasPaused flag propagate throughout this entire method. It means we have to think through every line change through both flows, install and resume

src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -495,7 +539,7 @@ export async function install(
 
     const installVersion = releaseVersion ?? gameInfoVersion ?? '0'
 
-    if (platformToInstall === 'Browser') {
+    if (platformToInstall === 'Browser' && !wasPaused) {
```
BrettCleary, line 542: I can't think of a scenario where we would be resuming an installation of a browser game since browser games aren't downloaded and install basically just sets the install url. This also goes back to my previous point about simplifying the code here

src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -642,6 +653,296 @@ export async function install(
...
+    process.noAsar = false
+
+    logInfo(`Error while extracting game ${error}`, LogPrefix.HyperPlay)
+
+    window.webContents.send('gameStatusUpdate', {
+      appName,
+      runner: 'hyperplay',
+      folder: destinationPath,
+      status: 'done'
+    })
```
BrettCleary, line 940: might want a `captureException` here too

BrettCleary (changes requested):
left a couple comments. looking good! thx for making the previously requested changes 🤝

src/backend/storeManagers/hyperplay/games.ts
red-game-dev: This was changed because upon resume download, and trying to continue the extraction does not happen anymore, I've tried various of methods, including calling within the `resumeIfPaused` with a lot of repeating logic, yet could not be executed. In this case, we had to change this method

Furthermore, this may change when we implement resume and pause for extraction

src/backend/storeManagers/hyperplay/games.ts
red-game-dev: Generally this should be called if after paused, that is why allowing only few places to let this go further, and calls the extraction later as intended

src/backend/storeManagers/hyperplay/games.ts
red-game-dev: Done, updated had to change quite of logic around

src/backend/storeManagers/hyperplay/games.ts
red-game-dev: Done

src/backend/__tests__/utils.test.ts
red-game-dev: Done

src/backend/__tests__/utils.test.ts
red-game-dev: Done

src/backend/storeManagers/hyperplay/games.ts
red-game-dev: Done

nyghtstalker:
1. Clicking the Play button after starting uninstall puts the the game into a broken state. It shows installed but unable to play or uninstall. This seems to be a Windows-only thing (I was not able to trigger this on Mac).
 
image
image

3. Text needs to be removed when extracting
image

nyghtstalker (approved):
The translation issue has been fixed. LGTM!

src/backend/services/ExtractZipService.ts
```diff
@@ -0,0 +1,400 @@
...
+    const elapsedTime = (Date.now() - this.#startTime) / 1000
+    const speed = this.#processedSize / elapsedTime
+
+    return {
+      progressPercentage,
+      speed,
+      totalSize: this.#totalSize,
+      processedSize: this.#processedSize
+    }
+  }
```
BrettCleary, line 193: might be good to round to nearest tenth of a percent here for `progressPercentage`. `elapsedTime` could be rounded to nearest second. not sure what units speed is but could round that too

Also we should rename these variables to include their units as it is only documented at the interface's comment and that should be documented on the variable name instead to keep the code readable

`elapsedTime` should be `elapsedTimeInSeconds`
`speed` should be `speedInMbPerSec` if those are the units
`totalSize` should be `totalSizeInBytes` or MB if those are the units. Same for `processedSize`

BrettCleary (changes requested):
looking good. left a couple comments. also I think you need to update your submodules to the latest main commits to fix some of the gh workflow errors

src/backend/services/ExtractZipService.ts
```diff
@@ -0,0 +1,400 @@
...
+      this.#onError(error as Error)
+      captureException(error)
+    } finally {
+      this.#zipFileInstance = null
+      this.#extractionPromise = null
+      this.#resolveExtraction = null
+      this.#rejectExtraction = null
+    }
+  }
+}
```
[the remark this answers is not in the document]
red-game-dev: Done

src/backend/services/ExtractZipService.ts
red-game-dev: Done

src/backend/services/ExtractZipService.ts
```diff
@@ -0,0 +1,419 @@
...
+import { open, ZipFile, Entry } from 'yauzl'
+import { mkdirSync, createWriteStream, rmSync, existsSync } from 'graceful-fs'
+import { captureException } from '@sentry/electron'
+import { join } from 'path'
+
+export interface ExtractZipProgressResponse {
+  /** Percentage of extraction progress. */
+  progressPercentage: number
+  /** Speed of extraction in bytes per second. */
+  speedInMbPerSec: number
```
BrettCleary, line 12: wait is this one bytes per second or megabytes per second?

BrettCleary (approved):
left one small comment

src/backend/services/ExtractZipService.ts
red-game-dev: Oh dam, was a typo. Cheers for noticing 🍻

red-game-dev: Updated

src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -134,10 +150,10 @@ export async function stop(appName: string): Promise<void> {
 
 export async function pause(appName: string): Promise<void> {
   const dl = inProgressDownloadsMap.get(appName)
-  if (dl === undefined) {
+  if (dl === undefined || dl.DownloadItem === undefined) {
```
flavioislima, line 153:
```suggestion
  if (!dl?.DownloadItem) {
```

Perhaps this is simpler?

src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -458,20 +503,61 @@ function getReleaseMeta(
...
-async function resumeIfPaused(appName: string): Promise<boolean> {
+async function resumeIfPaused(
+  appName: string
+): Promise<InstallResult | boolean> {
   const isPaused = inProgressDownloadsMap.has(appName)
+
   if (isPaused) {
     const item = inProgressDownloadsMap.get(appName)
-    if (item === undefined) {
+    if (item === undefined || item.DownloadItem === undefined) {
```
flavioislima, line 513:
```suggestion
    if (!tem?.DownloadItem) {
```

This one as well?

flavioislima (approved):
Pretty good stuff, man! :D

nyghtstalker:
Looked it over again and it looks good! Ship it!

PR #620, opened 2023-11-06
[FIX] Game Launch/Close Events

flavioislima:
While looking at our MixPanel and Rudderstack dashboard, we can almost not see any launch event at all and the launch conversion is pretty low. That makes no sense especially when compared with the other events, especially the install one.

The reason is that we were only tracking events for games that we could get the PID and this was failing in 99% of the cases. 
This PR adds the tracking event to the `launch` command so it can be more precise and catch all launch and close events no matter the platform or the source.

it also improves the Platform and Store names so they are more 'readable', instead of using the Runner only since it can be confusing to remember that Nile is Amazon and Legendary is Epic Games.

Also put all platforms into a main one, instead of two for Windows we have only one 'Windows' and so on.

Does not remove any code so the TrackPID is still there which will track the games twice in 1% of the cases which I think it is fine. But I can remove it if we won't use it anymore.

And I know in some cases it will start tracking the game while the dependencies are still being installed but that won't take more than 1 minute and only happens once on the first game launch.
It will also track the launcher of the game while it is opened but I also think this is not relevant because it is not the default behavior to just open the launcher and not open the game for several minutes or hours :P

image

image

---


eliobricenov (changes requested):
only this code style comment, the rest lgtm

src/backend/utils.ts
```diff
@@ -1575,3 +1576,38 @@ export const processIsClosed = async (pid: number) => {
     check()
   })
 }
+
+export const getStoreName = (runner: Runner) => {
```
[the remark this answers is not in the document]
flavioislima: ok bro. You can always make a suggestion as well :)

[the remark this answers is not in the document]
eliobricenov: oh rightttt my bad, I will update the status

eliobricenov (approved):

src/backend/metrics/metrics.ts
```diff
@@ -246,11 +252,12 @@ export async function trackPidPlaytime(
       trackEvent({
         event: 'Game Launched',
         properties: {
-          isBrowserGame: false,
+          isBrowserGame: gameInfo.browserUrl !== undefined,
```
red-game-dev, line 255: Maybe we could just have it `Boolean(gameInfo.browserUrl)` at least, since we need to consider if `browserUrl` could be empty str or such

flavioislima: I am actually not sure if we need that since we are passing the platform now. But I need to be sure that this is not being used by any dashboard on mixpanel.

red-game-dev: I guess we can just remove to be honest, because since we know the platform and we know the game to what belongs if is browse or not 🙌

flavioislima:
@BrettCleary waiting for you on any other parameters we can add, remove or change.
I just worried now that I changed the Store Names so this can break a few reports at least for a few weeks.
Also some fields might not being used or might not be necessary anymore. 
But no rush 👍🏽

src/backend/metrics/metrics.ts
flavioislima: Actually, this method is used only for native games so doesn't need any change. I reverted these.

src/backend/metrics/metrics.ts
```diff
@@ -268,8 +275,9 @@ export async function trackPidPlaytime(
           isBrowserGame: false,
           game_name: gameInfo.app_name,
           game_title: gameInfo.title,
-          store_name: gameInfo.runner,
+          store_name: getStoreName(gameInfo.runner),
```
flavioislima, line 278: Instead of `hyperplay`, `legendary`, `nile`... it will report:
`HyperPlay`, `Epic Games`, `Amazon Games`, etc.

Should I keep this change or it will cause some reports to break?

flavioislima: @awantoch

src/backend/metrics/metrics.ts
```diff
@@ -268,8 +275,9 @@ export async function trackPidPlaytime(
...
           processName: processInfo.name,
+          platform: getPlatformName(gameInfo.install.platform!),
```
flavioislima, line 280: same here. It won't send now `windows_amd64` only Windows, Linux, macOS or Browser.
Should I keep this change? @awantoch

BrettCleary: as a sidenote, I think we should condense all of our platform strings on game info into shared ones across all game stores (separate PR). Right now it's a leaky abstraction with different store naming conventions propagating everywhere

We should define a common interface for stores and enforce all stores to meet this standard. I think we probably only need "Windows, Linux, macOS or Browser" and the _amd64 etc can be inferred by the current arch

here's an issue I made for this https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/issues/627

src/backend/metrics/metrics.ts
awantoch: It should be pretty easy to compile the data from the old and new formats, I think the distinctive names are solid, so I def support the change

src/backend/metrics/metrics.ts
awantoch: ah yeah that makes sense -- Windows, Linux, macOS, Browser, but still probably helpful to include the architecture on the dev portal side, since we'll need to determine which version we run (mainly on macOS, determining whether to run the arm or amd version)

I still think it would be helpful to include the full platform_architecture combo in the log though for data purposes, if not just to report back to the developers 🤔

BrettCleary:
This breaks the game closed event for legendary games.

here you can see that it instantly sends closed after launched
image

This was also explained by my previous comment in src/backend/launcher

```
    /*
     * gogdl remains open while the game is running
     * by tracking gogdl instead of the game, we do not need to rely on
     * gogdl outputting the game's PID to stdout
     *
     * hyperplay and sideload game exes are launched directly so tracking the child process
     * works well unless the exe launches a separate process and then closes
     *
     * legendary closes after launching the game so we need to track the game's PID
     * which is outputted to stdout
     */
```

flavioislima:
> This breaks the game closed event for legendary games.
> 
> here you can see that it instantly sends closed after launched image
> 
> This was also explained by my previous comment in src/backend/launcher
> 
> ```
>     /*
>      * gogdl remains open while the game is running
>      * by tracking gogdl instead of the game, we do not need to rely on
>      * gogdl outputting the game's PID to stdout
>      *
>      * hyperplay and sideload game exes are launched directly so tracking the child process
>      * works well unless the exe launches a separate process and then closes
>      *
>      * legendary closes after launching the game so we need to track the game's PID
>      * which is outputted to stdout
>      */
> ```

Yeah, only on Windows though. Because legendary works that way. There is a PR to fix that on their end I guess, or at least an issue. because this breaks the autosync saves as well.
That is also why I did not removed the TrackPID method, it will be useful for Legendary games since HP will get the PID and use it to track the game. What is not working for HP games as we can see by the events on MixPanel.

And that is the whole point here, We don't really care about the other stores, we need HP game events being correctly tracked, right? And those are not being tracked and this is a HUGE issue since we use the number to trigger other things on the store for instance.

If you have a better solution, feel free to implement it but I think that tracking on exactly when the game is being actually launched works better, and you can try for yourself ofc.

BrettCleary:
The reason the launch events are failing for all native games is not because `child.pid` is undefined, it is because there was a regression on my cross platform overlay PR and `shouldTrackPlaytime` defaults to false

https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/pull/398/files#diff-7972f837982b2c43a359f127ede692926e5c9936df60856deb6f6cf17f3ecee0
file `src/backend/storeManagers/storeManagerCommon/games.ts`

Can we just set it to true on the callRunner call from storeManagerCommon? 

It makes no sense to have two systems for tracking game launched and game closed events. We should choose one and there are some pros and cons for each approach.

I chose to implement it at the child.spawn level for native games so that we could have this time played feature working for legendary launched games as well although this probably doesn't work after the heroic PR since I think that legendary version wasn't built from our fork. Maybe we should submit an upstream PR to output the PID or wait for that issue/PR you mentioned to go through.

If we want to support games with exe's that launch a separate process, we'll need to go back to requesting the developer to provide a processName or determinine it during the listing process. If we do this, we need the launch/closed at the PID level and to use the npm package that gets PID from process name, then poll until it launches.

At the ipcMain handle level like you are suggesting, it simplifies tracking across native and browser games though. Only need the trackEvents in that one ipcMain handler and we can remove the PID logic.

What do you think @flavioislima?

BrettCleary:
> We don't really care about the other stores

This is not true as we support forwarder listing to epic and we need metrics data for launch/closed for those games.

> I think that tracking on exactly when the game is being actually launched works better, and you can try for yourself ofc.

This is actually a point in favor of the existing approach at the child process spawn level

flavioislima:
@BrettCleary i don't really want to talk about numbers here since those PRs are public (it seems we are forgetting this here and getting too deep into how things work on our end).
But you can take a look at the number of installs from Epic and GOG. So I am just saying that we should not bother to have precise numbers for launch events for them because it is too inexpressive. 

About the solution, right now you know we have the Time tracking, right? It works fine and tracks the time the game was played well. So that is why I chose the same place to add the launch and close events. Because then we have a pretty good calculation about the play time and we can have also a precise number of the games people spend more time on. 
that place is also independent of the Store, platform, or anything and we have all the info we need there as well. 

So it is not only simple but pretty effective and, besides having these issues only on windows with epic games, it works perfectly for all other games, especially HP games.

BrettCleary (approved):
lgtm and fixes the game launched/closed events 🙌  we can probably remove the trackPIDPlaytime function and just use this approach since it's simpler

and then with the latest legendary changes, we'll have a fix for those legendary launched games

flavioislima:
Yeah, tracking PID would be a good solution If we didn't have to support so many platforms and different Runners. It is pretty hard to make it work with so many things involved.
But I think this more simple approach will work fine as well and we are not losing anything by changing to it :)

src/backend/metrics/metrics.ts
flavioislima: @awantoch so I keep as it is or use both then? we can have both, it is fine.

flavioislima: Added `platform_arch` to be used if we want to refine the architecture like you said 👍🏽

PR #622, opened 2023-11-06
[Fix] Game Update issues

flavioislima:
This PR fixes three issues:
- Canceling the update was deleting the game folder (it actually deletes when the update starts)
- Updating the game to the wrong folder (it was creating another folder with the game developer name and installing the game again on it, which would create a new folder level structure for every new update.)
- Update button on gamecard and submenu not working.

### To replicate the issue
- Start a game update and then cancel it. The game will show as not installed.
- And Finish an update and then check the game folder structure.

---


red-game-dev (approved):

src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -792,9 +790,12 @@ export async function update(appName: string): Promise<InstallResult> {
       accessCode = gameInfo.accessCodesCache[channelIdOfCurrentInstall]
   }
 
+  // two folders up from install path
+  const mainGameFolder = path.join(gameInfo.install.install_path, '../..')
```
BrettCleary, line 794: This will break updates for games installed on old versions of HyperPlay because the install_path was set to

```
    const sanitizedDestinationFolderName =
      developer !== undefined
        ? sanitizeFileName(developer) + ' - ' + sanitizeFileName(title)
        : sanitizeFileName(title)
```

so the game will be installed at the same level as the game folder and multiple games might be updated to be on top of one another in the same folder.

you can see that commit's code here https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/blame/d476f02f2d5d51497b1ff6de41903c15c6eef57f/src/backend/storeManagers/hyperplay/games.ts 

We should probably add tests to ensure against regressions with this backwards compatibility at some point

BrettCleary (changes requested):

src/backend/storeManagers/hyperplay/games.ts
flavioislima: The only thing I am doing is to set the right folder, should not break anything.
right now it is installing the game again but on a folder inside the game folder so you have a duplicated installation inside the current install and that is for every new update.

The only change I did was to put the files on top of the current ones, so it will extract and overwrite them.

So it basically does the same as before but one folder up.

In fact, I updated a game that was installed on a previous version of HyperPlay when testing this, because it was with my current install of bushi and Megaweapon.

BrettCleary: this breaks our backwards compatibility. It does not extract to the same folder if the game was installed with HyperPlay version <= 0.4.0 and is being updated with this branch. For those versions, games were installed at only 1 level below the games folder with the folder name I included in the code block above.

Therefore your code will go up two levels and extract the game to the same level as the games folder. It would look like this

Game installed on HP <= 0.4.0
```
GamesFolder
  * developer-title
    * game files
```

and this after updating with this branch
```
accountFolderName
  * projectFolderName
    * updatedGameFiles
GamesFolder
  * developer-title
    * game files
```

Perhaps we should add some code to detect this scenario and migrate the folder names first? The way we've been "migrating" has just been during update with the uninstall/install, but of course that's what we're trying to fix

BrettCleary: also we can remove the TODO on line 752 since that's what we're fixing

flavioislima: Yeah, that is complicated.
At the same time this is a pretty old version of HyperPlay that was release what? 3 or 4 months ago? 🤔 
Should we still think about supporting this old version?

i will try other things but since we do not have a way of knowing the hyperplay version that the game was installed and even the date of the installation, that will be complicated.

I think creating a whole new game folder with the all installed files is worse than dropping support for a old version 🤷🏽 

especially since on v0.4 we had just a few users compared to now.

flavioislima: Looking at the current install, megaweapon for instance is installed at:
`"install_path": "/home/flavio/Games/HyperPlay/benevolentmegacorp/megaweapon",

so if we use `install_path` it will reinstall the game on the megaweapon folder.
This means it will create a new `/benevolentmegacorp/megaweapon` folder inside it and won't delete any game file.

My solution gets the current install_path: and goes two folders up. Reinstalling the game on top of the current one.

But yes, if the game was installed on a folder called only megaweapon, this will install another game files. that is what is happening now with the current HyperPlay versions. 

How is the zip file btw? Does it contains the game files or it has this folder structure? if it has only the files then we can try to extract it directly into the current install no matter where it is.

BrettCleary: > How is the zip file btw? Does it contains the game files or it has this folder structure? if it has only the files then we can try to extract it directly into the current install no matter where it is.

The zip doesn't contain those `/benevolentmegacorp/megaweapon` folders. we create those on install

> Should we still think about supporting this old version?

Not sure. We should make a policy on this though.

If we want to keep support for those games, I think we could check if the folder already exists at /benevolentmegacorp/megaweapon and move the current install path to it prior to calling update. but it does add complexity and like you said this is a very old version and the games users have have probably been updated/migrated to the new folder structure since then. Honestly I'm fine to ignore it/have games installed on <= 0.4.0 update to the wrong folder.

BrettCleary (approved):
sounds good re our convo

looks like the TODO on line 752 is still in. you can remove that

flavioislima:
@BrettCleary added a new variable to the `install` command so it knows now if it is an update or not. This won't break installs on old versions of hyperplay.

So now it checks: if it is a new install, uses the method to get the path. If  it is an update, it extracts the zip to the current install folder. So no matter the folders structure it will work fine.

Thanks for the help :)

nyghtstalker (approved):
Great job fixing the updates via the game card. Those are now working. This now LGTM!

Issue #597, opened 2023-10-19
Add more in-depth events


awantoch:
We need to add the following events in the download manager:

- Install Paused
- Install Resumed
- Install Canceled

We also need to tweak the Install Failure event to not include pausing and canceling, as it conflates those currently.

The Game Launched event is also intermittently being emitted.

flavioislima:
The events can be seems on the screenshot from Rudderstack.
Also, as we can see on it, canceling the downloads sends a fail event as well. So maybe that is why we have so many fail events.


![Image](https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/assets/26871415/35d0dee3-36bd-4f6f-b78a-482467e65a1e)

PR #624, opened 2023-11-07
[Metrics] Add more Download Status Change Events

flavioislima:
Implements: #597 

This PR adds new events for change of installation status:
- Install Paused (both when the pause button is clicked and when canceling but keeping the files)
- Install Resumed 
- Install Canceled 

Image

---


flavioislima:
Am filtering out Fail installs that contain no error message since those are mostly because the user has canceled the install so we can have better tracking of the actual failure rate.

Image

src/backend/downloadmanager/utils.ts
```diff
@@ -132,6 +132,11 @@ async function installQueueElement(params: InstallParams): Promise<{
   }
 
   function trackFailedInstall(error: unknown) {
+    // undefined or 'Error' means the user aborted the install process
+    if (!error || error === 'Error') {
```
flavioislima, line 136: I believe that with no error message or just an 'Error' do not adds anything to our data analytics since it is not clear.

Other errors, for instance, when the URL is not reachable, will trigger an actual Error message that we can count as a fail.
Same with Permission errors or Disk Space errors.

eliobricenov: What about having a default error message "Unknown Error? It is the same as having a string with only "Error" but at least we'd see that it is happening. If we don't track the event, we have less visibility of error happenings.

eliobricenov (commented):
good stuff, just left a question

src/backend/downloadmanager/utils.ts
flavioislima: The thing here is that the error is not an error.
The whole problem is that we cannot pause is cancel the download without killing the process, there is no other approach to that. So it will always emit an error. But either silent or just with this Error message that happens to be the first string for any other error.

Having an unknown error would he just a renaming that requires actually more effort than we think of. Just try to figure it out real quick from where we trigger the error during the download. Might happens in too many places.

flavioislima:
I can also remove the part about ignoring the `Error` but then we need to change the dashboard on MixPanel and filter out the messages with only `Error` on it. I believe that ignoring the event is better since then we don't send events that are not important to us. 
@awantoch what are your thoughts on this?

src/backend/downloadmanager/utils.ts
BrettCleary: Hmm I'm going to have to disagree here. We are throwing these errors from `src\backend\utils\downloadFile\download_file.ts` for interrupted and canceled. We could add prefixes to those error messages and ignore them here if the error message matches.

If we're ignoring all exceptions without or with 'Error' error messages indiscriminately, we could accidentally ignore actual errors like Elio said. We could also refine error messages later for pause, interrupt, or cancel and now they are being tracked again.

Honestly I'm not sure we're even getting an error on pause with the new library. Instead of killing a process, it calls .pause on the download items from the map.

src/backend/metrics/types.ts
```diff
@@ -89,6 +89,33 @@ export interface GameInstallFailed {
...
+
+export interface GameInstallResumed {
+  event: 'Game Install Resumed'
+  properties: {
+    store_name: Runner
+    game_title: string
+  }
+  sensitiveProperties?: never
+}
+
```
BrettCleary, line 118: should we keep these properties at parity with the other game install events?

```
    store_name: Runner
    game_name: string
    game_title: string
    platform: InstallPlatform | AppPlatforms
 ```

File: src/frontend/components/UI/DownloadToastManager/index.tsx [hunk 3]
```diff
@@ -220,6 +221,28 @@ export default function DownloadToastManager() {
...
+    trackDownloadStatusChange({
+      event: 'Game Install Resumed',
+      properties: {
+        store_name: runner,
+        game_title: title
+      }
+    })
+    window.api.resumeCurrentDownload()
+  }
+
```
BrettCleary, line 245: Would this be easier to handle this in the main process?

BrettCleary (changes requested):
left some comments. I think we should handle resume, pause, cancel event tracking in the HyperPlay store. That way it's done in one place, testable, and we don't have to update these event calls while iterating on UI/renderer components.

src/backend/metrics/types.ts
flavioislima: This was my first approach but it is not possible to have reliable `platform` on GamePage and GameCard since they know nothing about the current installation besides the progress. So we would need to make them aware of that. So I decided to remove it and keep it more simple.

Are we using the appname for anything?
Also, the platform would be useful basically only if we decide to have a report like:
- Most paused games by platform.
- Most Resumed gamed by platform.
- Most canceled installation games by platform.
What do you think?

src/frontend/components/UI/DownloadToastManager/index.tsx
flavioislima: I decided to do it on the frontend because then we can track the actually interaction instead of what the backend processed. I believe that it makes more sense to have data about what the user wanted to do here instead of what how the app behave (since this we use Sentry already). But I am not data expert.

src/backend/downloadmanager/utils.ts
flavioislima: For HyperPlay games you are right about the Pausing behavior, it does not trigger any errors. But for the other stores they do.

Can you test pretty quick @BrettCleary what is the output when you cancel a game? Check if it shows 'Paused or Aborted', for me it does not show it anymore.

When I am canceling a HyperPlay game this is the whole output:

```(09:54:51) INFO:    [Backend]:          Aborting 0x799b758a631f05bbd441e865d0fc6be19a7b2884656c20abe5ca322bf4ebcb00
(09:54:51) INFO:    [Backend]:          Aborted 0x799b758a631f05bbd441e865d0fc6be19a7b2884656c20abe5ca322bf4ebcb00
(09:54:51) INFO:    [Backend]:          Canceled Bushi.zip
(09:54:51) INFO:    [Backend]:          Deleting abort controller for 0x799b758a631f05bbd441e865d0fc6be19a7b2884656c20abe5ca322bf4ebcb00
Cannot get game process name installedPlatform undefined or channelName undefined or gameInfo.channels [object Object] is undefined
(09:54:51) INFO:    [DownloadManager]:  0x799b758a631f05bbd441e865d0fc6be19a7b2884656c20abe5ca322bf4ebcb00 removed from download manager.
(09:54:51) INFO:    [DownloadManager]:  Removing /home/flavio/.config/hyperplay/.temp/0x799b758a631f05bbd441e865d0fc6be19a7b2884656c20abe5ca322bf4ebcb00...
(09:54:51) ERROR:   [Backend]:          Aborting not possible. Could not find a matching abort controller for 0x799b758a631f05bbd441e865d0fc6be19a7b2884656c20abe5ca322bf4ebcb00
(09:54:51) INFO:    [HyperPlay]:        Error while downloading and extracting game: undefined
(09:54:51) ERROR:   [Backend]:          Aborting not possible. Could not find a matching abort controller for 0x799b758a631f05bbd441e865d0fc6be19a7b2884656c20abe5ca322bf4ebcb00


(09:54:51) ERROR:   [DownloadManager]:  Installation of 0x799b758a631f05bbd441e865d0fc6be19a7b2884656c20abe5ca322bf4ebcb00 failed with: undefined
processNotification error


(09:54:51) WARNING: [DownloadManager]:  Installation of 0x799b758a631f05bbd441e865d0fc6be19a7b2884656c20abe5ca322bf4ebcb00 failed!
(09:54:51) INFO:    [DownloadManager]:  0x799b758a631f05bbd441e865d0fc6be19a7b2884656c20abe5ca322bf4ebcb00 added to download manager finished.
(09:54:51) INFO:    [DownloadManager]:  0x799b758a631f05bbd441e865d0fc6be19a7b2884656c20abe5ca322bf4ebcb00 removed from download manager.


(09:54:51) ERROR:   [DownloadManager]:  Installation of 0x799b758a631f05bbd441e865d0fc6be19a7b2884656c20abe5ca322bf4ebcb00 failed with: Error


(09:54:51) ERROR:   [DownloadManager]:  Installation of 0x799b758a631f05bbd441e865d0fc6be19a7b2884656c20abe5ca322bf4ebcb00 failed with: Error
processNotification error


(09:54:51) INFO:    [DownloadManager]:  Finished Installation of 0x799b758a631f05bbd441e865d0fc6be19a7b2884656c20abe5ca322bf4ebcb00
```

So we get TWO errors: `undefined` (that I am not sure it is going to MixPanel, but it is being sent to Rudderstack. And `Error` only, brings nothing to the table. 

Like I said, I am fine leaving it as it is tbh, it is just one more thing to filter on MIxpanel and two more events being sent to Rudderstack (where we have a limited number of events per month).

src/backend/metrics/types.ts
BrettCleary: technically the appName is the only truly unique identifier of the game. We could potentially have a game dev submit the same title for two separate game listings, which has come close to happening with The Wake for instance. 

Plus if we want to get the rest of the game's metadata, it will be simpler if we have the appName.

If we move these events to being called from the backend, I think we'd be able to send platform too. If it's too much of a hassle to get platform in that context, then sure we can ignore.

src/frontend/components/UI/DownloadToastManager/index.tsx
BrettCleary: I just really think this tracking should be in the main process (hyperplay store manager) with the rest of the logic and the frontend should just be for frontend state and react components (basically View Manager and View in MVVM model). Within that framework,  metrics logic is business logic that should be in the main process not in the view or view manager.

We will forget to add this when we add the pause button on the game page for instance. If it's in the main process, it is defined in one place and doesn't need updating. We can also mock this out and write tests more easily (and in one place) in the main process. It's more testable and maintainable.

If the app is not behaving in accordance with the user interaction, then our e2e or manual tests aren't working. We shouldn't incur tech debt to handle the case where that happens. We should just fix the functionality

Also do you think we'll expose pause, cancel, resume through the store or hp protocol sometime in the future? These metrics would not fire in that case if we leave them in the frontend

I'll let @red-game-dev and @eliobricenov weigh in here too

flavioislima: Ok, I will move it to the backend, no problem.
Although I don't agree that this is business logic and tracking what the user is doing makes more sense than the call per-se since we already have Sentry to catch issues on that.
But we need this merged so if we keep discussing here I don't think this will happen :P

src/backend/metrics/types.ts
flavioislima: Yeah, makes sense. I will re-add the appName.

src/backend/downloadmanager/utils.ts
flavioislima: I have created this new Board on Mixpanel since I am investigating the Download Errors. And we can see that the big majority of errors are coming as either `undefined` or  `Error`.
So I will change the code so this Error actually emits as 'Canceled' again and try to figure it out why the `undefined` ones.

in general, what is breaking the most is EPERM followed by Connection errors.

image

src/frontend/components/UI/DownloadToastManager/index.tsx
flavioislima: Done

src/backend/metrics/types.ts
flavioislima: done

src/backend/downloadmanager/utils.ts
flavioislima: I reverted this change to proper treat it on another PR

src/frontend/screens/DownloadManager/components/DownloadManagerItem/index.tsx
```diff
@@ -151,8 +152,24 @@ const DownloadManagerItem = observer(({ element, current, state }: Props) => {
...
     } else if (state === 'running') {
       window.api.pauseCurrentDownload()
+      trackDownloadStatusChange({
+        event: 'Game Install Paused',
+        properties: {
+          store_name: runner,
+          game_title: gameInfo.title,
+          game_name: gameInfo.app_name
+        }
+      })
```
BrettCleary, line 172: we're going to remove these now that we're tracking this in the download manager right?

src/frontend/screens/DownloadManager/helpers.ts
BrettCleary, line 1: and this

src/frontend/screens/Game/GamePage/index.tsx
```diff
@@ -890,8 +890,18 @@ export default observer(function GamePage(): JSX.Element | null {
...
     // resume download
     if (isPaused) {
+      trackDownloadStatusChange({
+        event: 'Game Install Resumed',
+        properties: {
+          store_name: runner,
+          game_title: gameInfo.title,
+          game_name: appName
+        }
+      })
```
BrettCleary, line 904: and this

src/frontend/screens/Library/components/GameCard/index.tsx
```diff
@@ -292,6 +293,30 @@ const GameCard = ({
...
+    trackDownloadStatusChange({
+      event: 'Game Install Resumed',
+      properties: {
+        store_name: runner,
+        game_title: gameInfo.title,
+        game_name: gameInfo.app_name
+      }
+    })
+  }
+
```
BrettCleary, line 319: and this

BrettCleary (changes requested):
If we're going to track this in the download manager, we need to remove the track event calls from the frontend right?

BrettCleary (approved):
awesome. thx for making the changes

flavioislima:
With so many branches I removed the trackpad from this one. Reverting...

PR #634, opened 2023-11-14
[FIX] MM Dropdown bug

### LINKED ISSUE #441: Dropdown bug
There is a bug in the dropdown to connect my wallet, it shows options that shouldn't show or should be enabled when I'm not connected to my wallet. We should improve the flow of these log-in options.

image

flavioislima:
FIX #441 

Not Connected:
image

Connected:
image

---


flavioislima:
I wonder if we should also use a different Text when not connected. Will change to `Connect Wallet` Let me know if you guys have any other suggestion.

flavioislima:
image

public/locales/en/translation.json
```diff
@@ -69,6 +69,10 @@
             "launchAborted": "Launch aborted",
             "moving": "Error Moving Game {{error}}",
             "title": "Error",
+            "ubisoft-connect": {
```
flavioislima, line 72: This is not related but was created when I ran `yarn i18n` Probably some leftover from some previous PR.

red-game-dev:
> I wonder if we should also use a different Text when not connected. Will change to `Connect Wallet` Let me know if you guys have any other suggestion.

Agreed, it should be `Connect Wallet` at least rather than `Not Connected`

red-game-dev (approved):

src/frontend/components/UI/AccountDropdown/index.tsx
```diff
@@ -1,40 +1,42 @@
...
       </NavLink>
     </Menu.Item>
   )
 }
 
-export default function AccountDropdown() {
+const WalletDropdown: React.FC = observer(() => {
   const { t } = useTranslation()
-  const { showMetaMaskBrowserSidebarLinks } = useContext(ContextProvider)
+  const showWalletLinks = walletStore.isConnected
```
BrettCleary, line 38: this will show the wallet links when MetaMask mobile/wallet connect wallets are connected. metamask snaps and metamask home should only be shown if MetaMask extension is connected

src/frontend/components/UI/AccountDropdown/index.tsx
```diff
@@ -1,40 +1,42 @@
 import { Menu } from '@mantine/core'
-import React, { useContext } from 'react'
+import React from 'react'
+import walletStore from 'frontend/state/WalletState'
```
BrettCleary, line 3: can we call it walletState? we have too many things called store

src/frontend/components/UI/AccountDropdown/index.tsx
```diff
@@ -50,48 +52,46 @@ export default function AccountDropdown() {
...
+            <NavLink
+              to={'/metamaskPortfolio'}
+              id={!showWalletLinks ? 'topElementWalletDropdown' : undefined}
+            >
+              <div className={`body ${styles.itemContents}`}>
+                {t('hyperplay.viewPortfolio', `View portfolio`)}
+              </div>
+            </NavLink>
+          </Menu.Item>
+        )}
```
BrettCleary, line 86: the portfolio should be shown when any wallet is connected, so you'll end up needing two flags 
```
showMetaMaskExtensionLinks = walletState.isConnected && walletState.provider === PROVIDERS.METAMASK_EXTENSION

showWalletConnectedLinks = walletState.isConnected
```

src/frontend/components/UI/AccountDropdown/index.tsx
```diff
@@ -50,48 +52,46 @@ export default function AccountDropdown() {
...
         <Menu.Item
           className={`${styles.menuItem} `}
           onClick={() => onboardingStore.openOnboarding()}
         >
           <div className={`body ${styles.itemContents}`}>
-            {t('hyperplay.changeWallet', `Change wallet`)}
+            {showWalletLinks
+              ? t('hyperplay.changeWallet', `Change wallet`)
+              : t('hyperplay.connectWallet', `Connect wallet`)}
```
BrettCleary, line 94: nice. I like this change 🥳

BrettCleary (changes requested):
nice. found a small issue that should be an easy fix

src/frontend/components/UI/AccountDropdown/index.tsx
flavioislima: sure, just copied paste from another component 😅

src/frontend/components/UI/AccountDropdown/index.tsx
flavioislima: so how should I fix that ?

src/frontend/components/UI/AccountDropdown/index.tsx
flavioislima: will fix

src/frontend/components/UI/AccountDropdown/index.tsx
BrettCleary: this comment is about the same issue https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/pull/634#discussion_r1394800406

src/frontend/components/UI/AccountDropdown/index.tsx
flavioislima: done

src/frontend/components/UI/AccountDropdown/index.tsx
flavioislima: Ok so now I changed it to only show MM portfolio and Snaps if it is MM wallet.
If is another wallet it will only show the Change Wallet button.

src/frontend/components/UI/AccountDropdown/index.tsx
flavioislima: should be fixed.

nyghtstalker:
@flavioislima I am getting this after connecting via MM mobile.

nyghtstalker:
> @nyghtstalker Can you test this behavior on `main` as well?
> 
I confirmed that it is also occurring in Main. That isn't good. The is def something wrong there. However, it doesn't seem like the PR is the cause.

> About the Change wallet text for the MM App, I think makes sense since you are connected with the mobile and want to connect with the extension, for instance. No?
> 
> What is the expected behavior in this case?

The problem is that the wallet IS connected (shows wallet address) but the menu is only displaying Change Wallet. Yes, the Change Wallet option should be in that menu along with all the other options.

flavioislima:
> 
> The problem is that the wallet IS connected (shows wallet address) but the menu is only displaying Change Wallet. Yes, the Change Wallet option should be in that menu along with all the other options.

But they work when connecting with the app? I am testing here on the stable and the portfolio does not work when connecting with the app. Does the snaps work? I enabled and disabled the options based on what interactions are working or not when using the app.

nyghtstalker:
> > The problem is that the wallet IS connected (shows wallet address) but the menu is only displaying Change Wallet. Yes, the Change Wallet option should be in that menu along with all the other options.
> 
> But they work when connecting with the app? I am testing here on the stable and the portfolio does not work when connecting with the app. Does the snaps work? I enabled and disabled the options based on what interactions are working or not when using the app.

The problem is when connecting with MM mobile. I know the mobile app has it's own quirks but once you get it to connect the user then gets that "Transaction Requested" popup (which is occurring in Main) and you can close this popup. However, the problem is the wallet menu. It only displays the one option - "Change Wallet". All of the other menu options are missing and that is the problem. 

With the wallet in this state, I confirmed that the wallet is missing in the full overlay. My guess is because the MM icon is missing.

image


Just found this - selecting Snaps in the menu does open the page, however, it also opens the wallet automatically.

1. Connect MM extenstion
2. Wallet menu -> View Snaps
3. Observe that wallet opens automatically

NOTE:
- Click the same View Snaps button does not trigger the wallet on this page
- "View fullscreen" or "View portfolio" does not trigger the wallet like "View Snaps" does

image

Connecting via MM extension seems to work correctly.

flavioislima:
@nyghtstalker 

So my question is: When connected with the **MM Mobile APP** should we show the link to:
- Portfolio?
- Snaps?
- MM Icon?
- Change Wallet?

This is what I can see on 0.9.0 when connected to the app:
image

No MM Icon and or Snaps link.

This is what happens when I click the View Portfolio:
image

But this is not related with this PR. The only thing my PR does is to change the links inside the submenu, I did not even touch the MM icon component or any other logic. So the other bugs, that I can reproduce as well, are all in the current release as should be fixed by another PR.

nyghtstalker:
> @nyghtstalker
> 
> So my question is: When connected with the **MM Mobile APP** should we show the link to:
> 
> * Portfolio?
> * Snaps?
> * MM Icon?
> * Change Wallet?
> 
> This is what I can see on 0.9.0 when connected to the app: image
> 
> No MM Icon and or Snaps link.
> 
> This is what happens when I click the View Portfolio: image
> 
> But this is not related with this PR. The only thing my PR does is to change the links inside the submenu, I did not even touch the MM icon component or any other logic. So the other bugs, that I can reproduce as well, are all in the current release as should be fixed by another PR.

I understand and agree with all that. My only concern is why we aren't seeing "View Portfolio" in the menu and only seeing "Change Wallet". I would think it should have the same functionality as Live. The other menu items are fine since they are not in Live. Everything else is fine.

nyghtstalker (approved):
After pulling the latest changes I am now seeing the correct menu for MMM that includes View portfolio.

BrettCleary (approved):

PR #649, opened 2023-11-27
Add token gated channels

jiyuu-jin:
Adds the ability to install from token gated channels from the desktop client InstallModal.


src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -458,6 +461,49 @@ async function getAccessCodeGatedPlatforms(
   return validateResult.platforms
 }
 
+async function getTokenGatedPlatforms(channel_id: number): Promise<PlatformsMetaInterface> {
+  if (!window.ethereum) throw 'Window.ethereum provider not found'
+  const provider = new ethers.BrowserProvider(window.ethereum)
+  const signer = await provider.getSigner()
```
red-game-dev, line 467: **Optional** but maybe we can have this as a helper function

```typescript
async function getSigner(): Promise<ethers.Signer> {
  if (!window.ethereum) {
    throw new Error('Ethereum provider not found');
  }
  const provider = new ethers.BrowserProvider(window.ethereum);
  return provider.getSigner();
}
```

src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -458,6 +461,49 @@ async function getAccessCodeGatedPlatforms(
...
+
+  const domain = window.location.host
+  const origin = window.location.origin
+
+  const statementRes = await axios.get(
+    valistBaseApiUrlv1 + '/license_contracts/validate/get-nonce'
+  )
+  const statement = String(statementRes?.data)
+
+  const siweMessage = new SiweMessage({
```
red-game-dev, line 478: **Optional**, but same as above we can have these as helper functions, this would make it easier if we need to utilize in multiple places

Can consider to chance the below `chain` as param too.

```typescript
async function createSiweMessage(signerAddress: string): Promise<SiweMessage> {
  const domain = window.location.host;
  const origin = window.location.origin;

  const statementRes = await axios.get(`${BASE_API_URL}/license_contracts/validate/get-nonce`);
  const statement = String(statementRes?.data);

  return new SiweMessage({
    domain,
    address: signerAddress,
    statement,
    uri: origin,
    version: '1',
    chainId: 1,
  });
}
```

Then we can easily use all this as 

```typescript
async function validateLicense(signer: ethers.Signer, channelId: number): Promise<PlatformsMetaInterface> {
  const address = await signer.getAddress();
  const siweMessage = await createSiweMessage(address);
  const message = siweMessage.prepareMessage();
  const signature = await signer.signMessage(message);

  const validateResult = (await axios.post<LicenseConfigValidateResult>(`${BASE_API_URL}/license_contracts/validate`, {
    message,
    signature,
    address,
    channel_id: channelId,
  })).data;

  if (validateResult.valid !== true) {
    throw new Error(`Address ${address} is not valid for channel id ${channelId}`);
  }

  if (!validateResult.platforms) {
    throw new Error('Token gated platforms returned by the validate url were undefined');
  }

  return validateResult.platforms;
}

async function getTokenGatedPlatforms(channelId: number): Promise<PlatformsMetaInterface> {
  const signer = await getSigner();
  return await validateLicense(signer, channelId);
}
```
You can move things accordingly as well, that is just an example the above

src/frontend/ExtensionManager/index.module.scss
```diff
@@ -1,11 +1,21 @@
...
+  left: 0px;
+  bottom: 0px;
   display: flex;
   flex-direction: row;
   gap: 20px;
   z-index: 100000;
   top: var(--top-navbar-height);
+  background-color: transparent;
+  border: none;
+  > div {
```
red-game-dev, line 14: I guess this because we cannot update the nested div with a class right? Since ideally we would have a class, as at this case it may go to each and every div nested

src/frontend/screens/Library/components/InstallModal/DownloadDialog/index.tsx
```diff
@@ -115,14 +117,17 @@ export default function DownloadDialog({
...
+  enableCTAButton,
+  requiresToken
 }: Props) {
   const previousProgress = JSON.parse(
     storage.getItem(appName) || '{}'
   ) as InstallProgress
   const { libraryStatus, platform, showDialogModal } =
     useContext(ContextProvider)
 
+  console.log({ requiresToken })
```
red-game-dev, line 129: It was only for debugging right?

src/backend/storeManagers/hyperplay/games.ts
jiyuu-jin: Awesome suggestion, I broke up the SIWE logic and moved some stuff around!

src/backend/storeManagers/hyperplay/games.ts
jiyuu-jin: Great idea, I broke this logic out into getSigner!

src/frontend/helpers/library.ts
```diff
@@ -90,6 +97,9 @@ async function install({
     storage.removeItem(appName)
   }
 
+  console.log('hello')
```
eliobricenov, line 100: some logs here

src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -618,8 +649,17 @@ export async function install(
       }
     }
 
-    // get presigned platform info if code gated
-    if (selectedChannel.license_config.access_codes) {
+    if (selectedChannel.license_config.tokens) {
+      console.log('I got here to token gating!')
```
eliobricenov, line 653: here too

eliobricenov (approved):
lgtm!

BrettCleary:
TODO: change valistBaseApiUrl back to the prod URL before merging

BrettCleary:
> TODO: change valistBaseApiUrl back to the prod URL before merging

done. ready for testing @nyghtstalker

src/backend/storeManagers/hyperplay/games.ts
BrettCleary: removed

src/frontend/helpers/library.ts
BrettCleary: removed

src/frontend/ExtensionManager/index.module.scss
BrettCleary: updated the selector

BrettCleary (approved):
LGTM. tested and is working!

nyghtstalker:
1. Clicking _anywhere_ in this top section (the links, the blank space, and the "X") seems to trigger the client to "relaunch" without closing the app - the app loads and the user is taken to the Store page like an initial app launch. A yellow box appears in the middle of the screen on this reload. 

image
 
2, Clicking "Sign and Install" has no effect when a wallet is not unlocked. We should add some sort of notification to the user since it seems like it is broken when nothing happens. Trigger a message to appear the Install button so it is easily noticed, popup telling the user...anything to inform them of the issue and how to fix it.

image

3. After signing the request install starts and then fails. Again, this can be confusing to the users and seems like it is broken. Getting this error in logs after.

image
image

nyghtstalker:
While testing with an **_artifact build_** and a wallet that DOES have the NFT:

1. With the wallet unlocked and on the Goerli network - clicking the "Sign and Install" button has no effect. No sign request gets sent, no flag on the MM icon, and nothing in the logs.

image

BrettCleary:
test failing is unrelated to this pr and is fixed here https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/pull/715

nyghtstalker:
Ccnfirmed that I am able to download the game with the wallet that has the NFT for Kosium. Below are a few concerns that need to be addressed.

1. I am getting this error in the logs after signing the request. I am guessing this is because this wallet _does not_ have an NFT. If this is the case I would recommend somehow _informing the user about the failure so they know what happened and how to correct the issue_. Right now from a user's perspective, this seems like the process is broken. 
```
(13:15:18) INFO:    [HyperPlay]:        Error while downloading and extracting game: Address code 0x350Cd4beCeF84E79411fe42ca45aaE67b71f4C5C is not valid for channel id 115!

(13:15:18) ERROR:   [Backend]:          Aborting not possible. Could not find a matching abort controller for 0xb5b77decf0bbeb40a2b9c5c85efc9e4dd72985fc9080733857bd7c2afc702f43

(13:15:18) ERROR:   [DownloadManager]:  Installation of 0xb5b77decf0bbeb40a2b9c5c85efc9e4dd72985fc9080733857bd7c2afc702f43 failed with: Address code 0x350Cd4beCeF84E79411fe42ca45aaE67b71f4C5C is not valid for channel id 115!
```

2. Yellow square is present when launching the app (using both dev mode AND the artifact build). Since it is occuring in both versions it is concerning that it will carry into live once it is released.

image

3. With the wallet LOCKED - clicking the "Sign and Install" button has no effect. This seems like it is broken. It should trigger the MM enter password window.

image

PR #651, opened 2023-11-28
[FIX] Back to previous position in library

### LINKED ISSUE #544: [Feature] Remember Scroll Position Library

flavioislima:
FIX #544 

https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/assets/26871415/bca0ef19-8100-46de-b509-992c1bf98d4c

### How to test

- On the library, scroll down and click on any game card. 
- Click on the 'Back' Button
- Should get back to the previous position

Might not work on small libraries with just a few games.

---


BrettCleary:
The library page now has nested scrolling on this branch
image


I can scroll the page and the listings independently. Just tested and confirmed this is not the case on main.

flavioislima:
>
> The library page now has nested scrolling on this branch

hmm. We need the overflow and the scroll if we want to get the scroll position. So the scroll is there but I set it to be invisible. Probably that is the cause of it. Perhaps if I change the scrolling to another element this could be fixed, but I tried on the Game List and it wasn't.
Our options are basically:
- Keep like this
- Show the scroll on the right for the Library Component
- Do not ship this fix and keep going back to the top when leaving the game page.

I will try a few more things.

> I am unable to leave the library page and return with scroll position remembered. it is always going back to the top for me (for both nested scrolls)

Weird, it is working fine for me. I will make a video.

flavioislima:
@BrettCleary fixed the issue with the nested scrolling. Moved the ref to the component above (contentContainer) instead of listing.

eliobricenov (approved):
code lgtm - tested and works fine

red-game-dev (approved):
Awesome! LGTM

BrettCleary (approved):
lgtm. thanks for clearing it up for me with the video!

PR #667, opened 2023-12-12
[Fix] WalletConnect Provider

BrettCleary:
Fixes WalletConnect connection. Previously, the connection would work and accounts would change correctly but sending any rpc request would fail.


red-game-dev (approved):
Awesome!

eliobricenov (approved):
just confirmed this works!

PR #674, opened 2023-12-13
[FEAT] Implement Feature Flags with LaunchDarkly

flavioislima:
Implements Electron and React SDK from [LaunchDarkly](launchdarkly.com).

This adds the ability to use Feature Flags for Enabling and disabling features without the need to make a new release. It also can be used for A/B/n tests for UX research and more.

### Changes

This PR per-se does not add any feature flag, I will add one in another PR.
LD provides SDKs for JS, Electron, and React. on my tests I did not need to use the JS one, we can add it in the future if we need to, but probably we won't since those already provide the abstractions we need.
I added a doc file with a few examples of how to load the flags on both FE and BE.

In Theory, we could ditch the React SDK as well since we can share the flags using IPC calls like I am doing here with the LD config. But I think that this can lead to some issues once we decide to disable a feature that is on both sides, causing the IPC call to not run and not update the result. So I think its safer to have both SDKs.

---


src/backend/main.ts
BrettCleary, line 1: could we put these changes in a separate module that main imports?

.github/workflows/flatpak-build.yml
```diff
@@ -46,12 +46,16 @@ jobs:
...
+      - name: setup env file
+        run: echo "VITE_LD_ENVIRONMENT_ID=${{ secrets.VITE_LD_ENVIRONMENT_ID_TEST }}" > .env
       - name: Install modules.
         run: yarn setup
       - name: Build artifacts.
         run: yarn dist:linux:ci:flatpak
       - name: Prepare Flatpak
         run: yarn flatpak:prepare
+      - name: setup env file
+        run: echo "VITE_LD_ENVIRONMENT_ID=${{ secrets.VITE_LD_ENVIRONMENT_ID_TEST }}" > .env
```
BrettCleary, line 58: is this needed twice in the same workflow?

.github/workflows/build.yml
```diff
@@ -37,6 +37,8 @@ jobs:
         run: echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" > ~/.npmrc
       - name: Install modules.
         run: yarn install --network-timeout 600000 && yarn allow-scripts
+      - name: setup env file
+        run: echo "VITE_LD_ENVIRONMENT_ID=${{ secrets.VITE_LD_ENVIRONMENT_ID_TEST }}" > .env
```
BrettCleary, line 41: we could make this a custom action under `.github/workflows/setupLD/action.yml`

then here in this step we would just set
```
- name: Setup Launch Darkly
        uses: ./.github/workflows/setupLD
```
 
 This way we don't have to duplicate the same 
```
  run: echo "VITE_LD_ENVIRONMENT_ID=${{ secrets.VITE_LD_ENVIRONMENT_ID_TEST }}" > .env
```
  everywhere and if we need to modify it in the future or add another LD setup step, we can do it in one place

BrettCleary (changes requested):
very cool! completely agree with keeping both sdk's. left a couple comments and questions

.github/workflows/flatpak-build.yml
flavioislima: no, what a mistake of copy/paste. Thanks for the catch

src/backend/main.ts
flavioislima: I think part of it is possible. The SDK initialization needs to be done once the app is ready though.

.github/workflows/build.yml
flavioislima: We would need two custom actions since one is for testing and other for production.

flavioislima: I tried it here but reverted in the end:
- I still needed to duplicate two lines of code on every action
- In some workflows the github cannot find the action
- After fixing the above, Github complains that actions has no access to secrets.

In the end, lost a lot of time for almost no benefit and reverted back to original.

eliobricenov (approved):
Changes LGTM!

After reading a bit the docs it seems they mention we could experience some UI flickery if we don't initialize the provider before the first render. Maybe we can benefit of awaiting to get the info before starting to render the app, an extra second could avoid layout shifts. This is not a blocker though, we can change this later if we decide we need it.

src/frontend/state/LDProvider.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+const LDProvider = ({ children }: Props): React.ReactElement => {
+  const [ldConfig, setLDConfig] = useState<LDEnv>({
+    envId: '',
+    ldUser: {
+      kind: 'user',
+      key: ''
+    }
+  })
+  const [provider, setProvider] = useState<React.ReactElement>(<></>)
```
eliobricenov, line 17: Have you tried what happens if we try to use `useFlags ` or `useLDClient` before the provider is initialized? We could get some errors because those hooks are expected to be used inside of an `<LDProvider />` but they initially are not.

flavioislima:
> Changes LGTM!
> 
> After reading a bit the docs it seems they mention we could experience some UI flickery if we don't initialize the provider before the first render. Maybe we can benefit of awaiting to get the info before starting to render the app, an extra second could avoid layout shifts. This is not a blocker though, we can change this later if we decide we need it.

We probably will need to adapt something once we have a real flag implemented, that will be my next PR. So then we can see if we will have flickering.

I think their docs are not updated with React 18 so I could not find a way of initializing it on the `index.tsx` without creating this `LDProvider` component. I tried first using an await call on the async method but when building the app Vite was complaining that 'top level awais was not supported'. So I tried updating our build to use `esnext` and did not work. Then I tried using a plugin for vite to add this features, and the build was failing as well. So I gave up and came with this component solution.

So I guess we need to test with a real flag like I said. But if you want to try something, let me know and feel free to send commits to this branch, no problem.

src/frontend/state/LDProvider.tsx
flavioislima: wdym? the LDProvider is being initialized with the App here:
https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/pull/674/files#diff-39a743b1ad7d31985313996d6fa54533b97d5ee693be0f992aca8ca2f653e7e4R114

So in theory, anything that uses these hooks on the app should work.

flavioislima: Another approach would be to use the JS SDK and load it on the head of the index.html, but we look the hooks.

eliobricenov: Oh no need, I feel we're on the right track with using hooks. 

What I mean is that from what I see`<LDProvider />` is initially `<></>` but after we load the config it becomes `<AsyncLDProvider/>`. In that period maybe using LD hooks could throw an error because they're not wrapped yet in the provider they need.

Either way I have not run the code locally so might be wrong.

eliobricenov:
> > Changes LGTM!
> > After reading a bit the docs it seems they mention we could experience some UI flickery if we don't initialize the provider before the first render. Maybe we can benefit of awaiting to get the info before starting to render the app, an extra second could avoid layout shifts. This is not a blocker though, we can change this later if we decide we need it.
> 
> We probably will need to adapt something once we have a real flag implemented, that will be my next PR. So then we can see if we will have flickering.
> 
> I think their docs are not updated with React 18 so I could not find a way of initializing it on the `index.tsx` without creating this `LDProvider` component. I tried first using an await call on the async method but when building the app Vite was complaining that 'top level awais was not supported'. So I tried updating our build to use `esnext` and did not work. Then I tried using a plugin for vite to add this features, and the build was failing as well. So I gave up and came with this component solution.
> 
> So I guess we need to test with a real flag like I said. But if you want to try something, let me know and feel free to send commits to this branch, no problem.

I can have a look tomorrow and if I see any improvements I'll create a PR. I'm super excited about having this in the app 🚀

flavioislima:
@eliobricenov check again, I tried another approach now and it seems to work fine on the `index.tsx`. I am using a async method to start the react process.

eliobricenov:
> @eliobricenov check again, I tried another approach now and it seems to work fine on the `index.tsx`. I am using a async method to start the react process.

Ohhh that's literally what I had in mind! Awesome, hopefully, it works. Thanks for spending some time trying an alternative!

flavioislima:
> > @eliobricenov check again, I tried another approach now and it seems to work fine on the `index.tsx`. I am using a async method to start the react process.
> 
> Ohhh that's literally what I had in mind! Awesome, hopefully, it works. Thanks for spending some time trying an alternative!

No problem, thanks for the tip. I mean, that could work but I think this new approach is safer since it will wait for the SDK always before start the app. The other case could have some delay like you said.

eliobricenov (approved):
amaizing stuff

BrettCleary (approved):
lgtm

red-game-dev (approved):
LGTM, amazing work! 🍾

PR #694, opened 2024-01-09
[FIX] Check for games updates on startup

flavioislima:
This fixes the issue with HP not checking for game updates on startup, leading users to play outdated games unless they hit the refresh button.

`checkUpdates` option is now `true` by default on all places and set to `false` only on the places we do not need it, like on a new login and browsing around.

### To test
- With games in need to update for any library, check if you can reproduce the issue on the stable release or main branch.
- When opening the library no game will show that it needs update, only after hitting the refresh button.
- Open the library again on this build and check if you can see the update icon on the game after the refresh finishes.

---

### DIFF
--- a/src/frontend/components/UI/ErrorComponent/index.tsx
+++ b/src/frontend/components/UI/ErrorComponent/index.tsx
@@ -21,7 +21,6 @@ export default function ErrorComponent({ message }: { message: string }) {
           className="button is-footer"
           onClick={async () =>
             libraryState.refreshLibrary({
-              checkForUpdates: true,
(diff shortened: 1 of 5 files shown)


BrettCleary (approved):
LGTM! i must have missed that when refactoring to mobx

eliobricenov (approved):

red-game-dev (approved):

nyghtstalker (approved):
Confirmed that restarting the client will perform a refresh and will now prompt a user to update if needed.

PR #699, opened 2024-01-11
[FEAT] Offline improvements

flavioislima:
This PR improves some offline handling:

- Redirect to the Library instead of the Store when open without internet (currently, it shows only a blank page on startup).
- If navigating on any store and then lose the internet, it also gets redirected to the library.
- If offline the Store button gets disabled
- Fixes the Offline Message bar since now it was totally black because of z-index issues.

imagem

---


flavioislima:
I noticed that you can import and create a new wallet from MM when offline, although at the end it tries to redirect and it never finishes loading. The wallet is created though.

Do you guys think we should handle that as well and maybe not initiate MM if offline?

red-game-dev (approved):
LGTM 🍾

red-game-dev:
> I noticed that you can import and create a new wallet from MM when offline, although at the end it tries to redirect and it never finishes loading. The wallet is created though.
> 
> Do you guys think we should handle that as well and maybe not initiate MM if offline?

This shouldn't take much time and not much changes? In that can be handled, let me know so I can recheck the PR in case 🙌

flavioislima:
> > I noticed that you can import and create a new wallet from MM when offline, although at the end it tries to redirect and it never finishes loading. The wallet is created though.
> > Do you guys think we should handle that as well and maybe not initiate MM if offline?
> 
> This shouldn't take much time and not much changes? In that can be handled, let me know so I can recheck the PR in case 🙌

I think the effort is not much, the main thing here is that if we want to users be able to setup a wallet when offline.
Because the wallet can be used normally for offline transactions I guess, haven't tested with a proper game.

Another thing I did not try was to see if the proxy works while offline 🤔 
also, we can keep improving things when offline, we will still discover new issues or UX improvements especially for people who uses it on the Steam Deck.

red-game-dev:
> > > I noticed that you can import and create a new wallet from MM when offline, although at the end it tries to redirect and it never finishes loading. The wallet is created though.
> > > Do you guys think we should handle that as well and maybe not initiate MM if offline?
> > 
> > 
> > This shouldn't take much time and not much changes? In that can be handled, let me know so I can recheck the PR in case 🙌
> 
> I think the effort is not much, the main thing here is that if we want to users be able to setup a wallet when offline. Because the wallet can be used normally for offline transactions I guess, haven't tested with a proper game.
> 
> Another thing I did not try was to see if the proxy works while offline 🤔 also, we can keep improving things when offline, we will still discover new issues or UX improvements especially for people who uses it on the Steam Deck.

Understood, true I think this is a question we should discuss through dev meeting, it is more ideal for everyone's input in this case

eliobricenov (approved):

BrettCleary:
> I noticed that you can import and create a new wallet from MM when offline, although at the end it tries to redirect and it never finishes loading. The wallet is created though.
> 
> Do you guys think we should handle that as well and maybe not initiate MM if offline?

We need to allow MM in offline mode. Like you said, wallet creation and import still works and we want to encourage that.

It can still be useful in that state if we ever get an offline game as well. Take for instance a player playing an offline game that uses a wallet signature in the game to create their account (only verified client side in this example or perhaps it will sync saved game data after some playing and when back online). 

If the user is able to import/create MM account, then they can still proceed with this without internet.

There is also the case where they are playing a multiplayer game on a local network server. We would still want them to be able to sign messages to sign into those games even if they are only connected locally. 

Another case could be connecting MM to a local testnet. 

So there are a lot of use cases for using MM in offline mode even if they are edge cases.

BrettCleary (approved):
Did we want to style/design this page on this PR or another?

nyghtstalker:
1. Blank screen still displays if the internet is lost while the app is open. The changes from this PR do not take effect until the app is launched offline.

image

2. Once the internet is restarted while the app is open a loading wheel appears and appears not to stop. You can scroll down to find the loaded store. However, a black section is now stuck at the top of the page. This persists in the store game pages too.

image
image
image
image

- Leaving the store and returning does fix this issue.

image

3. When the app is launched while offline the library page has a black section at the top added and the MM menu gets cut off at the top.

image

flavioislima:
Good catches, Joe.
The Offline bar on top was already broken for a while now, I added a fix to it:
imagem

I also changed the logic a bit.
Now, instead of rendering a message saying that it's offline, the app will navigate away from the store if the internet is lost, it will disable the Store icon and show a message saying that no internet connection was found.

It is easier and more intuitive this way.

So if the internet is back the button gets enabled again.

flavioislima:
Also, use the build from the PR instead of loading in dev mode while testing the PR. Somehow the hot-reload gets weird when toggling offline/online while the dev server is running.

nyghtstalker:
**FYI - This testing is using the latest artifact build and not dev mode.**

1. While the client is launched WITH internet and remains open -> disconnect internet -> go to Library and then back to Store -> Blank page

image

2. While the client is launched WITH internet and remains open -> go to Store page -> disconnect internet -> Store page does not change while internet is disconnected

image


3. While the client is launched WITH internet and remains open -> disconnect internet -> go to Library and click Refresh button -> Loading wheel will endless spin -> reconnect internet -> return to Store and endless load remains -> click Refresh button -> endless load still remains

image

4. This error is triggered when the client is launched WITHOUT internet

image



On a positive note - when the client is launched without internet the Store icon is now grayed out, inaccessible, and displays a message on hover. This is awesome!

image

flavioislima:
@nyghtstalker 

1.  I cannot replicate this since it starts on the library when offline.
2. On my end it redirects to the library a few seconds after getting offline. Video Below
[Screencast_20240130_144723.webm](https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/assets/26871415/37b39d59-547d-4772-ae17-80f56c6bcb00)

3. Checking
4. Fixed on latest commit
5. Yes, added the message after it. Check the PR description again, I updated it and the changes I introduced.

flavioislima:
There are probably other bugs when offline but I have no intention on fixing ALL of them on this PR, only the most prominent ones when the app is opened. In the future we can fix more.

nyghtstalker:
@flavioislima - I am still seeing issues 1-2 occurring. Confirmed that these are working on Linux. This may be a Windows issue that needs to be invested more. The MM menu is being cut off again as I have mentioned previously.

image

flavioislima:
> @flavioislima - I am still seeing issues 1-2 occurring. Confirmed that these are working on Linux. This may be a Windows issue that needs to be invested more. The MM menu is being cut off again as I have mentioned previously.
> 
> image
> 

Tested here on windows and I cannot replicate the issues.  Needed to film it with my phone since windows was failing to record the screen: 

https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/assets/26871415/9db11e97-43be-4abd-a0ea-e18f1ed993cd

flavioislima:
Fixed the MM extension on the latest commit:
imagem

nyghtstalker:
@flavioislima That is really weird that it is still working for you when using Windows! Are you using a VM? I just tried it again with your latest changes and I am still having the same results. I got a video for you this time. What is weird is when I launch without internet and turn it back on then it WILL detect once the internet is restored and reconnects. It seems like the "check for the internet" is not working when I launch WITH INTERNET CONNECTED and never detects when it goes offline. 

https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/assets/82125535/68060d38-d891-408e-85e0-6c6fd0567b2b

flavioislima:
> @flavioislima That is really weird that it is still working for you when using Windows! Are you using a VM? I just tried it again with your latest changes and I am still having the same results. I got a video for you this time. What is weird is when I launch without internet and turn it back on then it WILL detect once the internet is restored and reconnects. It seems like the "check for the internet" is not working when I launch WITH INTERNET CONNECTED and never detects when it goes offline. 
> 
> https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/assets/82125535/68060d38-d891-408e-85e0-6c6fd0567b2b
> 
> 

Yeah, it is weird. I'm not using a VM, it is a stand windows 11 installation.

Tested again on a vm and worked fine as well 🤷

Anyway, I don't think this is a blocker though.
What do you think?

nyghtstalker:
> > @flavioislima That is really weird that it is still working for you when using Windows! Are you using a VM? I just tried it again with your latest changes and I am still having the same results. I got a video for you this time. What is weird is when I launch without internet and turn it back on then it WILL detect once the internet is restored and reconnects. It seems like the "check for the internet" is not working when I launch WITH INTERNET CONNECTED and never detects when it goes offline.
> > 
> >   
> >     
> >     
> > 
> >     OfflineModeNotWorkingMyWindows.mp4
> 
> Yeah, it is weird. I'm not using a VM, it is a stand windows 11 installation.
> 
> Tested again on a vm and worked fine as well 🤷
> 
> Anyway, I don't think this is a blocker though. What do you think?

No, I think it is fine. It is a bummer that we aren't certain but we will deal with it if it becomes a problem.

nyghtstalker (approved):

PR #716, opened 2024-01-29
[FIX] HP Games import

flavioislima:
- Fix current logic
- Use Manifest file to get version, platform and exe path
- Check for subfolders if dev folders are selected
- Update manifest after import with new info
- Show error messages if invalid folders or EXE not found.

## How to test

**IMPORTANT: Should be tested on all operating systems and for native and non-native games (Windows games on macOS and Linux)
It is not possible to import zipped games, only extracted ones.
It is not possible to import Sideloaded games. Only the ones that were previously installed from the HP store.
The game should be added to the library first, it is not possible to install a game that it is not on the library.
The changes affect only HP games, not Epic and GOG which are working fine and have their logic.**

### Testing if Import works
0. Not necessary but can improve testing speed: Get a game you already have installed, go to the game folder, rename it, or copy to a different place. Then, you can go ahead and uninstall the game from HP. 
1. Go to the install modal and click IMPORT. 
2. Select the Dev folder (the folder that contains the game folder and not the game files).  The game should be installed now and works fine after clicking Select.
3. Repeat Step 1 but now select the Game folder (the one with the game files that is inside the dev folder). The game should be installed and works fine after selecting it.

### Testing the error messages

1. Invalid folder:
     * For any non-installed game, click Install > Import. Select any folder on your machine that does not contain the game. For instance: C:\username\Picures.
     *  An error should be shown saying that the folder is invalid.
   
2. Missing executables:
      * Select the proper game folder, but the game executable is missing
      * An error will show that it is not possible to import since the game files are missing

---


src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -157,6 +161,36 @@ export async function pause(appName: string): Promise<void> {
   dl.DownloadItem.pause()
 }
 
+// check for valid json file inside the folder before importing
+// if none was found, check the first folder inside the folder
+// return the path to the folder with a valid json file
+// if none was found return empty string
+// this is necessary because we do not know which folder the user has selected, the main folder or the game folder
+const isValidGameFolder = (appName: string, folderPath: string): string => {
```
eliobricenov, line 169: this is nit pick and nonblocker: I find it weird that this function doesn't return a boolean even though it uses a `is__` prefix.

BrettCleary: I agree. Seems `getValidGameFolderPath` is more accurate

flavioislima: Good point.
I think that was the idea but then I decided to return the folder path instead.
I will rename it.

red-game-dev (approved):
LGTM 🍾

BrettCleary:
@flavioislima found an issue. might have been introduced in the previous PR but it is keeping me from testing this one

I 
1. downloaded Kosium
2. renamed it to kosium_2
3. Saw that the game card was still shown as available but the game wouldn't launch of course (silent failure)
4. On app relaunch I see a manifest file is created in the kosium folder and the game card still shows as available

image
image

flavioislima:
> @flavioislima found an issue. might have been introduced in the previous PR but it is keeping me from testing this one
> 
> I
> 
>     1. downloaded Kosium
> 
>     2. renamed it to kosium_2
> 
>     3. Saw that the game card was still shown as available but the game wouldn't launch of course (silent failure)
> 
>     4. On app relaunch I see a manifest file is created in the kosium folder and the game card still shows as available
> 
> 
> image image

Question: did you refresh the library after renaming the folder? or relaunched HP? or changed to game page?
Because the method to check if the game is available is only triggered on library load/reload/navigating to/from gamepage/library.
We dont have a listener to the filesystem changes because that would be insane on huge libraries I believe. 
So if you did not did that then it is expected.
Also the json created makes sense since it could not find on the game folder.

nyghtstalker (approved):
Awesome quality of life improvement! Well done! This will come in handy during my testing! =D

PR #742, opened 2024-02-22
[UX] Use download folder install path for .temp

BrettCleary:
Download into a .temp folder in the default install path by default. If the default install path setting is not set, use the config folder as was used previously

---


red-game-dev (approved):

nyghtstalker (approved):
Confirmed that a new hyperplay folder is created with a .temp folder inside of it.

image

image

PR #744, opened 2024-02-22
[UX] Install to .temp in game destination directory

BrettCleary:
Move the download install directory inside the destination directory so all downloads and extractions happen on the same directory for accurate space requirements estimates

---


nyghtstalker (approved):
Confirmed that the .temp folder gets installed in the same folder as the .exe.

image

image

PR #745, opened 2024-02-22
[Windows] Allow more install options

flavioislima:
On Windows:

- Allow to select a different install directory
- Allow install for all users as administrator
- Disable one click install (will always show prompt)
- Always delete appdata on uninstall 

---


red-game-dev (approved):

BrettCleary (approved):

nyghtstalker:
- Confirmed I am able to install and launch HP from a different drive.

image

- It looks like all appdata gets removed (Settings and Library). Going back to the live build my library and settings are wiped. **_I feel this should be retained. Is there a specific benefit to wiping this?_**

image

image
image

- It looks like the wallet has been reset. This may be because the wallet folder name is a hash of the file path of the hyperplay install location and since this is in a new location it changes.

image

- There is an uninstaller now.

image


- Extra stores are displayed on the Manage Stores page. **_We need to make sure these will remain hidden after merging._**

image

- My EPIC store library is still displayed but the EPIC tab is missing.

image

- Unable to launch EPIC games

image

flavioislima:
@nyghtstalker 
- The issue with the epic game is that we store the login information on the app data but the game listing is on .config/legendary. We can change that since Heroic did some time ago but its not in the scope of this PR. So its expected that it will ask to login again. So just login again and it will fix.
- The wallet reset for several reasons, but if the appdata was removed I believe that its expected that it will ask to configure again.
- Amazon is showing because the feature flag is enabled in test. It is disabled in prod.
- Appdata was removed because you uninstalled HP and this is the behavior now in prod, no?
- The filter disappeared because it uses the login information that is on the appdata for that. It is a bug for sure but I think not on the scope of this PR to fix. The games should not be loaded if the user is not logged in.

Ping me if you have more questions 💪🏽

nyghtstalker (approved):
This is good. User can install at a location of their choice.

PR #574, opened 2023-10-05
[CI] Add mac arm64 runner

BrettCleary:
Build arm64 mac on a m1 runner to fix bytecode error

This uses a manual trigger to save $. Once it is merged main you will be able to manually trigger for any branch here https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/actions/workflows/build-mac-arm64.yml

### Testing
1. Verify the error from 0.7.1 on m1
2. Download `mac-dmg-arm64` https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/actions/runs/6424604877?pr=574
3. Sign the app with `xattr -r -d com.apple.quarantine /Applications/HyperPlay.app`
4. Verify the app launches without error

---

Use the following Checklist if you have changed something on the Backend or Frontend:

- [ ] Tested the feature and it's working on a current and clean install.
- [ ] Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)
- [ ] Created / Updated Tests (If necessary)
- [ ] Created / Updated documentation (If necessary)

PR #746, opened 2024-02-22
[CI] Fix macOS release workflow

flavioislima:
Fix the issue with the intel release overwriting the arm64 one by keeping one latest-mac.yml file only.

---


.github/workflows/build-mac-arm64.yml
```diff
@@ -1,45 +0,0 @@
...
-  workflow_dispatch:
-
-env:
-  GITHUB_TOKEN: ${{ secrets.WORKFLOW_TOKEN }}
-  GH_TOKEN: ${{ secrets.WORKFLOW_TOKEN }}
-  RUDDERSTACK_CLIENT_ID: ${{ secrets.RUDDERSTACK_CLIENT_ID }}
-
-jobs:
-  release-mac-arm64:
-    runs-on: macos-latest-xlarge
```
BrettCleary, line 13: we need this m1 runner because we compile to bytecode and the bytecode has to be compiled on the same arch that it will be distributed for https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/pull/574

.github/workflows/build.yml
```diff
@@ -108,12 +108,18 @@ jobs:
...
-          name: mac-dmg-x64
+          name: mac-intel
           path: dist/HyperPlay*x64.dmg
           retention-days: 7
+      - name: Upload ARM build.
+        uses: actions/upload-artifact@v3
+        with:
+          name: mac-arm64
+          path: dist/HyperPlay*arm64.dmg
+          retention-days: 7
```
BrettCleary, line 125: we can't compile arm builds on `macos-latest`

.github/workflows/release_macOS.yml
```diff
@@ -83,6 +48,6 @@ jobs:
         run: echo "VITE_LD_ENVIRONMENT_ID=${{ secrets.VITE_LD_ENVIRONMENT_ID }}" > .env.production
 
       - name: Build artifacts.
-        run: yarn release:mac:arm64
+        run: yarn release:mac
```
BrettCleary, line 51: I don't think we can compile intel builds on this arm runner `macos-latest-xlarge` because of the bytecode issue https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/pull/574

package.json
BrettCleary, line 1: need these commands separated to address the bytecode errors fixed on https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/pull/574

BrettCleary (changes requested):
This would be a regression of https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/pull/574

.github/workflows/release_macOS.yml
flavioislima: I think we should try since it is being 5 months since the change.
I will modify the other file to see how it will work.

PR #748, opened 2024-02-22
Fix: Resolved an issue where multiple profiles were not displayed if the sequence of numbers was inconsistent.

red-game-dev:
This description should serve as a reference in case we need near the future to explain these changes and as well for us to know what caused this issue, as per our investigation.

The issue encountered was affecting the visibility of user browser profiles. specifically, when profile sequences are not consecutive (e.g., missing Profile 3 in a sequence of Profile 1, Profile 2, Profile 4), our system would prematurely halt the display of profiles beyond the last consecutively numbered profile found (Profile 2 in this scenario). This behavior resulted in the omission of subsequent profiles, regardless of their presence or correct configuration.

However, profiles containing an improperly installed or absent MetaMask extension would not be displayed, leading to incomplete user experiences. Users need to ensure their MetaMask extension is correctly installed to avoid such display issues.

The attached screenshot illustrates a case where, before implementing the fix, only profiles 1 through 6 would be displayed, contingent upon the availability of MetaMask in those profiles

image

## Steps to Reproduce & Test

- Have or Create at least 3 Chrome Profiles. You can add them from the 3 dots top-right corner -> Your Profile Icon & name -> Add Profile. Each profile requires to be signed in to a different Google account, and it cannot be a guest as you will need to install extensions.
- Install MetaMask extension on each, and sign in
- Go to Explorer at `C:\Users\<Your Username>\AppData\Local\Google\Chrome\User Data` there you will find multiple profiles, such as `Profile 1`, `Profile 2`, `Profile 3`. On the `Profile 2` either rename it to `Profile 5` or some other number higher than your current or delete it (Only if it is the new profile). Please do not remember to not delete your day-to-day profile as you will lose your extensions and worse (Especially if you do not have a copy of your metamask recovery seed). If you rename it it should still appear, if you delete it, it should not appear but yet still able to see `Profile 3` in that case.
- Go to the client `yarn setup && yarn start` 
- Go to Connect Wallet
- Go to Connect using the MetaMask extension
- Go to Chrome for this case and you should be able to see all profiles that are eligible.

FYI: If the extension is not installed on one of the Chrome Profiles, you won't be able to view it, nor if is not installed correctly.
 
QASA: https://app.qase.io/case/HC-327

---


BrettCleary (approved):
awesome work. let's let @nyghtstalker test first before merge

BrettCleary:
also need to fix lint @red-game-dev

nyghtstalker (approved):
Confirmed that I am seeing multiple profiles on the extension selection screen and I am able to connect and unlock the wallet in HP. After resetting HP I was able to connect a different extension with a different wallet.

image

image

image

eliobricenov (approved):

PR #761, opened 2024-02-26
[UX] Improve Compatibility layer setup and handling

flavioislima:
This PR does several things like:

- Adds a warning on the install dialog when installing a non-native game so the person is aware that might have issues or the game not even work
- Adds a warning when launching non-native games as well. The 'Dont show again' button stores the option per game, not globally. This is intentional since most people won't check if the game is native or not and the result also depends on the game, hardware, system, etc.
- Adds a way of checking if macOS is updated to Sonoma or later so we can enable or disable GPTK automatically.
- Adds a check to see if Rosetta is installed on mac M1+ and instructions on how to install it if not.
- Use Wine-crossover if macOS ir not updated.
- If GPTK is selected to run a game, show a warning about it and an option to use Wine instead.
- Even if GPTK is available, if not on latest macOS it will download Wine.
- Fix Winetricks download being skipped on macOS and also uses main URL for it.

### How to test

This one is tricky to test if you have updated your Mac to Sonoma I believe.
But the screen below should appear for all Windows games on both macOS and Linux:
The other changes would require me to change the logic to check for a version of Mac that does not exist yet or with a Mac using Ventura or below.

Warning on Installing non-native game:
Screenshot 2024-02-26 at 20 22 01

Warning when launching non-native game:
Screenshot 2024-02-26 at 20 38 45

Warning on Rosetta not available:
Screenshot 2024-02-27 at 10 51 09

---


flavioislima:
The lint check was failing but it was related with the previous PR. That is why there are some changes on some non-related files.

public/locales/en/gamepage.json
flavioislima, line 1: Linting, for some reason the JSON was using different spacing than weblate on all i18n files.

src/backend/main.ts
```diff
@@ -296,12 +298,21 @@ async function initializeWindow(): Promise<BrowserWindow> {
...
+      (wine) => wine.type === 'toolkit'
+    )
+    const shouldDownloadWine =
+      !availableWine.length || (isMac && !toolkitListDownloaded)
+
     Promise.all([
       DXVK.getLatest(),
       Winetricks.download(),
-      !availableWine.length ? downloadDefaultWine() : null
+      shouldDownloadWine ? downloadDefaultWine() : null
```
flavioislima, line 311: Although we check for GPTK here, on the downloadDefaultWine we will skip GPTK and download wine for non-compatible macOS.

src/backend/tools.ts
```diff
@@ -393,15 +393,12 @@ export const DXVK = {
 
 export const Winetricks = {
   download: async () => {
-    if (!isLinux) {
+    if (isWindows) {
```
flavioislima, line 396: This was why Winetricks was broken on macOS. it was never being downloaded.

src/backend/tools.ts
```diff
@@ -393,15 +393,12 @@ export const DXVK = {
...
       return
     }
 
-    const linuxUrl =
+    const url =
```
flavioislima, line 400: Also the standard winetricks works fine on mac, no need for a different one than linux.

src/backend/utils.ts
```diff
@@ -899,6 +899,26 @@ async function ContinueWithFoundWine(
   selectedWine: string,
   foundWine: string
 ): Promise<{ response: number }> {
+  const isGPTK = selectedWine.toLowerCase().includes('toolkit')
```
flavioislima, line 902: The check here for toolkit only is intentional. If I want to avoid all toolkits, even from third-party like Whisky, I would need to add more stuff. But since those other GPTK might be modified and even run on older versions, I think we should try.
this is definitely something that only pretty advanced macOS users would notice anyway and they probably knows their ways around it.

src/backend/utils.ts
```diff
@@ -915,19 +935,35 @@ async function ContinueWithFoundWine(
...
   // use Wine-GE type if on Linux and Wine-Crossover if on Mac
-  const release = availableWine.filter((version) => {
+  const release = availableWine.filter(async (version) => {
     if (isLinux) {
       return (
         version.type === 'Wine-GE' && version.version.includes('Wine-GE-Proton')
       )
     } else if (isMac) {
-      return version.type === 'Game-Porting-Toolkit'
+      const isGPTKCompatible = await isMacSonomaOrHigher()
```
flavioislima, line 961: Here we decide which Compatbiility layer will be the default based on the macOS version. Wine-Crossover for old ones and GPTK for Sonoma and later.

src/backend/utils.ts
```diff
@@ -969,14 +1005,47 @@ export async function downloadDefaultWine() {
   return null
 }
 
+export async function setGPTKDefaultOnMacOS() {
```
flavioislima, line 1008: If compatible with GPTK we also want to setup a new prefix as default to avoid issues with DXVK installed on a previous Wine prefix.
The downside is for games that stores saves locally, they will be 'lost' so will need to be moved from the previous prefix to a new one.
But this is a global config, not individual, so it is unlikely to happen.

src/backend/utils.ts
```diff
@@ -969,14 +1005,47 @@ export async function downloadDefaultWine() {
...
   logFileLocation: string
 ): Promise<boolean> {
   const wineIsValid = await validWine(gameSettings.wineVersion)
 
-  if (wineIsValid) {
+  const isToolkit = gameSettings.wineVersion.type === 'toolkit'
+  const isGPTKCompatible = await isMacSonomaOrHigher()
+
+  const isValidOnLinux = isLinux && wineIsValid
+  const isValidOnMac =
```
flavioislima, line 1044: Things got a bit convoluted here but I am not sure if there is a better way to write this code.

src/frontend/components/UI/DialogHandler/components/MessageBoxModal/index.tsx
```diff
@@ -72,7 +75,15 @@ const MessageBoxModal: React.FC<MessageBoxModalProps> = function (props) {
       <DialogHeader onClose={props.onClose}>{props.title}</DialogHeader>
       <DialogContent className="body dialogContent">
         {getContent()}
+        {props.showCheckbox && (
```
flavioislima, line 78: This will show a warning for games launched using compatibility layer.
It includes a 'Do not show again' button. This is stored individually for each game so the person will see this warning when trying a new windows game on mac or linux.

src/backend/utils.ts
```diff
@@ -915,19 +936,79 @@ async function ContinueWithFoundWine(
   return { response }
 }
 
+export async function isRosettaAvailable() {
```
eliobricenov, line 939: non-blocker: return false instead of undefined

src/backend/utils.ts
```diff
@@ -915,19 +936,79 @@ async function ContinueWithFoundWine(
...
+  const result = rosettaCheck.split(':')[1].trim() === '1'
+
+  logInfo(
+    `Rosetta is ${result ? 'available' : 'not available'} on this system.`,
+    LogPrefix.Backend
+  )
+
+  if (!result) {
+    // show a dialog saying that hyperplay wont run without rosetta and add information on how to install it
+    await dialog.showMessageBox({
```
eliobricenov, line 965: This looks odd, we're showing a warning dialog in a function that checks if rosseta is available.
Can we move it outside?

src/backend/utils.ts
```diff
@@ -969,14 +1050,47 @@ export async function downloadDefaultWine() {
...
+  const isGPTKCompatible = await isMacSonomaOrHigher()
+  if (!isGPTKCompatible) {
+    return
+  }
+
+  const wineList = await GlobalConfig.get().getAlternativeWine()
+  const gptk = wineList.find((wine) => wine.type === 'toolkit')
+  if (!gptk) {
+    await downloadDefaultWine()
+    return
```
eliobricenov, line 1063: how are we retrying after we early return here

src/backend/utils.ts
```diff
@@ -969,14 +1050,47 @@ export async function downloadDefaultWine() {
...
+    logInfo(`Changing wine version to ${gptk.name}`)
+    GlobalConfig.get().setSetting('wineVersion', gptk)
+    // update prefix to use the new one as well
+    const installPath = GlobalConfig.get().getSettings().defaultInstallPath
+    const newPrefix = join(installPath, 'Prefixes', 'GPTK')
+    GlobalConfig.get().setSetting('winePrefix', newPrefix)
+
+    return true
+  }
+  return false
```
eliobricenov, line 1075: these booleans are to check the function returned successfully?

src/backend/utils.ts
BrettCleary: could maybe break these up into two variables `isMac && isToolkit && isGPTKCompatible && wineIsValid` and `isMac && !isToolkit && wineIsValid`. I'm not sure what you'd name them. could add a comment too if it still seems unclear

src/backend/utils.ts
flavioislima: Yeah, true. I will think on how to do that, maybe just rename the method.
Because actually, we are not using the return here for anything, its just to check and show the warning.

src/backend/utils.ts
flavioislima: no, I will fix it. This method is only to adjust the prefix, not really to return anything.

src/backend/utils.ts
flavioislima: Good catch, I will call this method itself after finishing downloading Default Wine.

src/backend/utils.ts
flavioislima: done

src/backend/utils.ts
flavioislima: removed it and its only void now

src/backend/utils.ts
flavioislima: changed the name of the method

BrettCleary:
> Adds a warning when launching non-native games as well. The 'Dont show again' button stores the option per game, not globally. This is intentional since most people won't check if the game is native or not and the result also depends on the game, hardware, system, etc. 

I was unable to trigger this one

BrettCleary:
Found a bug:

1. Try to launch the game 
2. Modal warning shows
3. exit out of modal
4. Try to launch game again
5. Game will not launch. silent failure

If I navigate to a separate screen and back, I can then go through the same flow again @flavioislima

BrettCleary:
> If GPTK is selected to run a game, show a warning about it and an option to use Wine instead.
I was unable to trigger this one. Are there other requirements to trigger this dialog?

BrettCleary (changes requested):
code looks good. requesting changes for the two bugs I found

flavioislima:
> > If GPTK is selected to run a game, show a warning about it and an option to use Wine instead.
> 
> I was unable to trigger this one. Are there other requirements to trigger this dialog?

yes, you need to be on macOS 13 or lower.
If you have Sonoma then it wont be triggered.

flavioislima:
> Found a bug:
> 
> 1. Try to launch the game
> 2. Modal warning shows
> 3. exit out of modal
> 4. Try to launch game again
> 5. Game will not launch. silent failure
> 
> If I navigate to a separate screen and back, I can then go through the same flow again @flavioislima

can you check if you have any Wine or GPTK installed?
Which macOS version do you have?

flavioislima:
> > Adds a warning when launching non-native games as well. The 'Dont show again' button stores the option per game, not globally. This is intentional since most people won't check if the game is native or not and the result also depends on the game, hardware, system, etc.
> 
> I was unable to trigger this one
> 
> Update: This one will not show if you do not have a wallet connected. The onboarding modal seems to take precedence

hmm, for me it shows the wallet modal and then the other one. 🤔

flavioislima:
> > Adds a warning when launching non-native games as well. The 'Dont show again' button stores the option per game, not globally. This is intentional since most people won't check if the game is native or not and the result also depends on the game, hardware, system, etc.
> 
> I was unable to trigger this one
> 
> Update: This one will not show if you do not have a wallet connected. The onboarding modal seems to take precedence


https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/assets/26871415/449d861a-182b-469a-af73-76c23dffcd81

BrettCleary (approved):
retested and it is now working great for me

nyghtstalker (approved):
Loving all the added messages to the user! Great job!

image

image

PR #763, opened 2024-02-26
feat: add email subscription modal

eliobricenov:

The modal will show after the user finishes the onboarding.

How to test:
- import a MM wallet or connect MM mobile or wallet connect

---


BrettCleary (approved):

src/frontend/screens/Onboarding/walletSelection/index.tsx
```diff
@@ -144,6 +145,7 @@ const WalletSelection: React.FC<WalletSelectionProps> = function (props) {
     dbPath?: string,
     browser?: ImportableBrowser
   ) {
+    authState.enableOpenAuthModalOnAppReload()
```
eliobricenov, line 148: enable user imports MM and reloads the app

src/frontend/screens/Onboarding/walletSelection/index.tsx
```diff
@@ -165,6 +167,7 @@ const WalletSelection: React.FC<WalletSelectionProps> = function (props) {
     })
     wait(4000).then(() => {
       props.disableOnboarding('connected')
+      authState.openSignInModal()
```
eliobricenov, line 170: open after user finishes onboarding

red-game-dev (approved):

eliobricenov:
A good portion of the component is similar to the `AuthModal` component but I wanted to decouple them so we have more control over it. By using a different electron partition we also make sure that we get the same layout even if the user logs in, which is not the case if we use the `AuthModal`.

nyghtstalker (approved):
Works great!

BrettCleary:
works great for me

let's add an .env.example file with the LD flag

BrettCleary:
Tested again and it works great for me!

PR #765, opened 2024-02-27
Add auto launch HyperPlay option

BrettCleary:
# Summary
Add auto launch HyperPlay option. default to true

# Testing
- [X] Windows
- [ ] Linux
- [x] Mac


red-game-dev (approved):
LGTM

src/backend/utils/auto_launch.ts
```diff
@@ -0,0 +1,35 @@
+import AutoLaunch from 'auto-launch'
+import { GlobalConfig } from 'backend/config'
+import { ipcMain } from 'electron'
+
+export const hyperPlayAutoLauncher = new AutoLaunch({
+  name: 'HyperPlay',
+  path: process.execPath
+})
+
+export async function updateAutoLaunch(enable: boolean) {
```
eliobricenov, line 10: nitpick: can we make the argument named to make it easier to see we're enabling or disabling it?

```ts
function  updateAutoLaunch(mode: 'enabled' | 'disabled') {
  if (mode === 'enabled')  { ... } 
  else { ... }
}

updateAutoLaunch('enabled');
updateAutoLaucnh('disabled');
```

eliobricenov (approved):

src/backend/utils/auto_launch.ts
BrettCleary: I think boolean makes more sense here since it is a toggle option with only two states 

I also would have to change the param, the call, the settings type, etc. for this. Not sure if it's worth it

PR #771, opened 2024-02-28
feat: send game update notifications

eliobricenov:
How to test:
- Have in the library a game with an available update or change the `library.json` file in the app data.
- Run the app
- You should see the notifications pop up
---


src/backend/storeManagers/index.ts
```diff
@@ -91,6 +94,64 @@ export function autoUpdate(runner: Runner, gamesToUpdate: string[]) {
...
+
+    gamesToUpdate.push(...gamesWithRunner)
+  }
+
+  if (gamesToUpdate.length === 0) {
+    return
+  }
+
+  const { runner, game } = gamesToUpdate[0]
+  const leadGameInfo = gameManagerMap[runner].getGameInfo(game)
```
BrettCleary, line 129: I think we should try to use a HyperPlay game as the leadGameInfo if possible

src/backend/storeManagers/index.ts
```diff
@@ -91,6 +94,64 @@ export function autoUpdate(runner: Runner, gamesToUpdate: string[]) {
...
+    runner: string
+    game: string
+  }[] = []
+
+  for (const runner in libraryManagerMap) {
+    const games = await libraryManagerMap[runner].listUpdateableGames()
+
+    const gameSettings = await Promise.all(
+      games.map(async (game) => gameManagerMap[runner].getSettings(game))
+    )
```
BrettCleary, line 108: this could take a long time for libraries with 100's of games and it looks like you're only using the first game info below and testing if >1 game to update. 

I also think we should maybe just get HyperPlay games info since those are the games we want to advertise the most

Could you do a filter for HyperPlay games first and then just map the first two results?

BrettCleary (changes requested):

src/backend/storeManagers/index.ts
eliobricenov: perfect that saves a lot of computation

BrettCleary:
working great 
image

BrettCleary (approved):

src/backend/main.ts
```diff
@@ -810,6 +811,13 @@ ipcMain.handle('checkGameUpdates', async (): Promise<string[]> => {
     oldGames = [...oldGames, ...gamesToUpdate]
   }
 
+  sendGameUpdatesNotifications().catch((e) =>
+    logError(
+      `Something went wrong sending update notifications: ${e}`,
+      LogPrefix.Backend
+    )
+  )
```
BrettCleary, line 819: this will run on every refresh click. shouldn't we only run this once on startup? we can add a module level flag in the `src/backend/storeManagers/index.ts` module for this

BrettCleary: fixed this

nyghtstalker:
I was unable to see any notification during this testing. I could be the exception. I created a ticket to investigate this more.
https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/issues/773

PR #774, opened 2024-02-28
[Fix] Prevent wine install on windows

BrettCleary:
Wine was being installed on Windows, making the app slow to launch

---


flavioislima (approved):
Good catch. I was able to replicate the issue on windows on the new release. Probably introduced on my improve compatibility layer PR.

I think only adding a check on download default wine should fix it but it's good to add those checks as well 👍

PR #777, opened 2024-02-28
[Summon Quest] Add LD flag, error handling, metrics events

BrettCleary:
This was giving `fetch failed` on HyperPlay startup and in general does not need to be active until the Summon quest is live


eliobricenov (approved):

flavioislima (approved):
Lgtm

PR #778, opened 2024-02-28
[FIX] Flathub release workflow

flavioislima:
Fix the current issues:
- Workflow failed to add release notes and open the PR because the structure of the original XML changed due to Flathub made it mandatory to add a Copyright comment at the top. So I adjusted the position of the `ul` element to the correct one.
- Moved from xml2js over to fast-XML-parse that is the most updated lib and doesn't mess with formatting.

Tested it running the command locally and the result can be observed on this PR:
https://github.com/flathub/xyz.hyperplay.HyperPlay/pull/80

---

### DIFF
--- a/flathub/update-flathub.ts
+++ b/flathub/update-flathub.ts
@@ -101,7 +101,7 @@ async function main() {
   )
 
   const releaseNotesJson = JSON.parse(releaseNotesStdOut)
-  const releaseNotesComponents = releaseNotesJson.body.split('\n')
+  const releaseNotesComponents: string[] = releaseNotesJson.body.split('\n')
 
   const hpXmlJson = convert.xml2js(hpXml, { compact: false }) as Element
   const releaseNotesElements: Element[] = []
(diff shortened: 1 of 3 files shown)


eliobricenov (approved):

flavioislima:
Rewrote the code to use a different library because the previous flathub PR was failing. Now it passed and I merged it, everything works fine.

flathub/update-flathub.ts
```diff
@@ -101,43 +100,69 @@ async function main() {
...
-  hpXml = convert.js2xml(hpXmlJson)
-  fs.writeFileSync(xmlFilePath, hpXml)
+
+  releaseListTag.ul = [...releaseNotesElements] // Set the <ul> element to the generated <li> elements
+
+  console.log('new releaseListTag = ', JSON.stringify(releaseListTag, null, 2))
+
+  const updatedHpXml = builder.build(hpXmlJson)
+
+  console.log('updatedHpXml = ', updatedHpXml)
```
BrettCleary, line 163: just double checking if you meant to leave the console logs in

BrettCleary (approved):
doing God's work sir 🙏

flathub/update-flathub.ts
flavioislima: Yes, good to debug in case we have future issues like the previous one.
Also causes no harm 😬

PR #783, opened 2024-03-05
[Fix] EBUSY during download cancel

BrettCleary:
Removing the directory when canceling an in progress download was being called twice and failing with EBUSY. Selecting to keep the files would still attempt to delete them.

TBD how to handle cancel during extraction with Yes/No to keeping download files.

### DIFF
--- a/src/backend/api/downloadmanager.ts
+++ b/src/backend/api/downloadmanager.ts
@@ -84,8 +84,8 @@ export const handleDMQueueInformation = (
 export const cancelDownload = (removeDownloaded: boolean) =>
   ipcRenderer.send('cancelDownload', removeDownloaded)
 
-export const cancelExtraction = (appName: string) =>
-  ipcRenderer.send('cancelExtraction', appName)
+export const cancelExtraction = (appName: string, removeDownloaded: boolean) =>
+  ipcRenderer.send('cancelExtraction', appName, removeDownloaded)
 
 export const resumeCurrentDownload = () =>
   ipcRenderer.send('resumeCurrentDownload')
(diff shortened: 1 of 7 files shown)


src/backend/api/downloadmanager.ts
```diff
@@ -84,8 +84,8 @@ export const handleDMQueueInformation = (
 export const cancelDownload = (removeDownloaded: boolean) =>
   ipcRenderer.send('cancelDownload', removeDownloaded)
 
-export const cancelExtraction = (appName: string) =>
-  ipcRenderer.send('cancelExtraction', appName)
+export const cancelExtraction = (appName: string, removeDownloaded: boolean) =>
```
eliobricenov, line 87: I think we should pass an object instead of a flag.

```ts
canceExtraction = ({ appName, removeDownloaded }: {appName: string; removeDownloaded: boolean }) => {...}
```

src/backend/utils/rmFolderRecursive.ts
```diff
@@ -0,0 +1,42 @@
+import fs from 'fs'
+
+export function deleteFolderRecursive(path: string) {
```
eliobricenov, line 3: why not using rmSync with recursive flag?

```ts
fs.rmSync(dir, { recursive: true, force: true });
```

BrettCleary: doesn't work unfortunately. only works if the folder is already empty

tried all the solutions from https://stackoverflow.com/questions/18052762/remove-directory-which-is-not-empty and this is the only one that worked

flavioislima:
@BrettCleary What is the current state of this PR? Does it still need to be fixed?

BrettCleary:
closing as stale

PR #784, opened 2024-03-05
[FEAT] Show System information on Log Screen

flavioislima:
This pulls Heroic's System information component and adds it to the Log Settings Screen. 

Should help a lot with support tickets.

I think we should not have a section only for that imo since we have a lot of tabs already and it is related.

Windows:
telegram-cloud-photo-size-1-5132234277042367614-y

macOS:
image

Linux:
image

---


flavioislima:
Since we don't have a design for it for now I will keep Heroic's, but will open a ticket so designers can add it to our Design system and on another PR we can improve it.

src/backend/utils/inet/downloader/index.ts
BrettCleary, line 1: is inet short for internet here? do we need the extra folder?

src/backend/utils/inet/downloader/index.ts
```diff
@@ -0,0 +1,102 @@
...
+        }
+      }
+    }
+  }
+
+  const response = await axios
+    .get(url, {
+      ...options.axiosConfig,
+      headers: {
+        'User-Agent': `HeroicGamesLauncher/${app.getVersion()}`,
```
BrettCleary, line 74: We should change this user agent to HyperPlay

src/backend/utils/inet/downloader/index.ts
BrettCleary: I understand this is from Heroic, but this is a pretty big module just to add for fetching and caching `https://pci-ids.ucw.cz/v2.2/pci.ids`. We already have a download file module at `src/backend/utils/downloadFile/download_file.ts` too. 

The save to file cache logic seems like overkill, but I can see how it could help if the request fails often.

I'm fine with leaving it but I just wonder if there is a simpler way to do the caching and revalidation and to reuse what we already have

BrettCleary (changes requested):
requesting changes for the user agent.

This is really a lot of code (~500 lines) to add and duplicate across repositories that we will have to maintain. I think we should really strongly consider moving the backend pieces into an npm package in the Heroic Github org adding that package to hp. 

The backend functionality to get the detailed system info seems like it could be broken off pretty cleanly from the rest of the codebase.

I could see other projects wanting to use the work the Heroic team as done here as well.

src/backend/utils/inet/downloader/index.ts
BrettCleary: if we move to a separate npm package, then disregard my comment here

flavioislima:
> 
> This is really a lot of code (~500 lines) to add and duplicate across repositories that we will have to maintain. I think we should really strongly consider moving the backend pieces into an npm package in the Heroic Github org adding that package to hp.
> 
> The backend functionality to get the detailed system info seems like it could be broken off pretty cleanly from the rest of the codebase.
> 
> I could see other projects wanting to use the work the Heroic team as done here as well.

I agree that this is too much of a sophisticated solution.
It doesn't need too much maintenance though so I am not sure about having an NPM package, although I agree that would remove a lot of code from our codebase. Just need to think about the effort x value of moving everything to a new lib and update code on both apps and test it.

I could also just use the same info we have on the logs and show there on top of the logs, we would just lose some details because of not using the pciids though.

src/backend/utils/inet/downloader/index.ts
flavioislima: >
>The save to file cache logic seems like overkill, but I can see how it could help if the request fails often.
>

Yeah, it used to be simple but it was failing and making the app crash or freeze when entering the settings screen. 
Another issue after that was that we were making a lot of requests to the pci-ids website so their mantainer opened an issue for us to solve that and we needed to increase the caching period.

So this version here already contains those fixes.

BrettCleary (approved):
tested and it is working for me

red-game-dev (approved):

nyghtstalker:
Loving the added System Info!!! 

- Windows 
image

nyghtstalker (approved):

PR #786, opened 2024-03-05
fix: wallet intermittently resets and restarts onboarding

eliobricenov:
### Summary

**We need to https://github.com/HyperPlay-Gaming/hyperplay-extension-helper/pull/41 to be merged**

Currently, we're using the `isInitialized` flag from the [extension-helper store](https://github.com/HyperPlay-Gaming/hyperplay-extension-helper/blob/0dfa36a915a35947aef1eaec00cd110b1a4de632/ipcHandlers/index.ts#L325) to decide what we should show when users click the MM browser extension button in the [onboarding modal](https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/blob/main/src/frontend/screens/Onboarding/walletSelection/index.tsx#L105).

The value is persisted in the default config file (`~/Library/Application Support/hyperplay/config.json`), but it is removed when the app is updated. I haven't figured out where yet, but the value of the config file is overwritten during the installation and reset to the default values. By moving the extensions helper store to its own config file, we can prevent the extension helper config from being overwritten. Each store should have its own file anyway, and the extension helper was not using the `TypeCheckedStoreBackend` class, so I think it is a good update.

**Solves https://github.com/HyperPlay-Gaming/product-management/issues/177**

How to test:

1. Build the app with `yarn dist:mac:arm64` or `yarn dist:mac:x64` depending on your machine
2. Import your browser wallet
3. Close app
4. Bump the package.json version `0.12.0` -> `0.13.0`
5. Build the app again
6. Install the newest version of the app
7. Check that the already imported wallet is detected

With this we emulate the app being updated. 

I've only tested this with Mac though, but we should give it a go in other platforms.

---

### DIFF
--- a/src/common/types/electron_store.ts
+++ b/src/common/types/electron_store.ts
@@ -106,6 +106,9 @@ export interface StoreStructure {
   providerMetadata: {
     currentWeb3Provider: PROVIDERS
   }
+  extensionMetadata: {
+    isInitialized: boolean
+  }
 }
 
 export type StoreOptions<T extends Record<string, unknown>> = Store.Options<T>


eliobricenov:
**We need to https://github.com/HyperPlay-Gaming/hyperplay-extension-helper/pull/41 to be merged**

BrettCleary (approved):
lgtm. let's have @nyghtstalker test

red-game-dev (approved):

nyghtstalker (approved):
Confirmed that my imported wallet remains after changing the version for all platforms. Good job!

PR #829, opened 2024-03-31
[Quests] Client integration

BrettCleary:


src/frontend/hooks/useGetQuest.ts
```diff
@@ -0,0 +1,26 @@
...
+      if (questId === null) {
+        return {}
+      }
+      const response = await window.api.getQuest(questId)
+      if (!response) return null
+      return response
+    },
+    {
+      refetchOnWindowFocus: false
+    }
```
eliobricenov, line 18: You might want to make this a dependent query: https://tanstack.com/query/latest/docs/framework/react/guides/dependent-queries, and only enable it if `questId` is defined.

src/frontend/hooks/useGetQuest.ts
```diff
@@ -0,0 +1,26 @@
...
+  const queryKey = `getQuest:${questId}`
+  const query = useQuery(
+    queryKey,
+    async () => {
+      if (questId === null) {
+        return {}
+      }
+      const response = await window.api.getQuest(questId)
+      if (!response) return null
+      return response
```
eliobricenov, line 14: if you add a type casting here you can remove the `ts-ignores` in the QuestViewer components

```ts
return response as {
    name: string
    capsule_image: string
}
```

src/frontend/components/UI/QuestsViewer/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+      ? t('quest.claimAll', 'Claim all')
+      : t('quest.connectSteamAccount', 'Connect Steam account'),
+    questType: {
+      REPUTATION: t('quest.reputation', 'Reputation')
+    }
+  }
+
+  if (
+    selectedQuestId !== null &&
+    Object.hasOwn(questResult?.data?.data ?? {}, 'name')
```
eliobricenov, line 100: this looks off, why do you need to do this? can typing the query help remove this?

eliobricenov (approved):
great stuff. I left some minor comments, feel free to address them in a different PR

red-game-dev (approved):

src/frontend/hooks/useGetQuest.ts
BrettCleary: Good idea. I think you meant to add this comment to the `useGetSteamGame` hook. I added it there. I also made both fields optional as technically this is an `unknown` type since we don't have runtime type checking on the http response

src/frontend/components/UI/QuestsViewer/index.tsx
BrettCleary: Yeah. Just removed it. I was having issues with the keys/re-rendering at one point and wanted to ensure that each component had a unique key. was using this to debug

This http response is really an unknown type though. maybe we can add zod to the ipcmain handler at some point

src/frontend/hooks/useGetQuest.ts
BrettCleary: made the change! didn't know about this feature

nyghtstalker (approved):
This looks AWESOME! I love it! I can't wait to see this fully loaded and ready to go! Bad ass! Rockin job!!!

PR #830, opened 2024-04-01
[CI] Use Native Notarise method from Electron builder

flavioislima:
Moving away from `electron/notarise` since there are bugs where it fails to notarize more than one file (for instance arm and x64). Also not needed since Electron builder latest version has a builtin module for that.

Tested these changes on Heroic since the macOS release was broken there.

---


src/frontend/screens/Login/index.tsx
```diff
@@ -2,7 +2,7 @@ import React, { useContext, useEffect, useState } from 'react'
 import './index.css'
 import Runner from './components/Runner'
 import { useTranslation } from 'react-i18next'
-import { useNavigate } from 'react-router'
+import { useNavigate } from 'react-router-dom'
```
flavioislima, line 5: unrelated to the PR but those imports were wrong for a long time now so I don't think we need a PR for that alone.

src/frontend/screens/WebView/index.tsx
```diff
@@ -6,7 +6,7 @@ import React, {
   useState
 } from 'react'
 import { useTranslation } from 'react-i18next'
-import { useNavigate, useLocation, useParams } from 'react-router'
+import { useNavigate, useLocation, useParams } from 'react-router-dom'
```
flavioislima, line 9: same here

red-game-dev (approved):
LGTM 🙌

eliobricenov (approved):

PR #831, opened 2024-04-01
[FIX] Check disk space not working

flavioislima:
Fix an issue introduced after migrating to Vite5 where the check-disk-space library was not working correctly. So the calculation was failing on the install modal.

---


eliobricenov (approved):

nyghtstalker (approved):
Disk space is now displaying correctly.

image

PR #832, opened 2024-04-02
[Feat] Get access code on update

BrettCleary:


nyghtstalker (approved):
Confirmed that it will ask you to update with a code. Confirmed that the previous code to install it will not work - a new one is needed for the update.

red-game-dev (approved):

src/frontend/state/UpdateState.ts
```diff
@@ -0,0 +1,73 @@
+import { makeAutoObservable, runInAction } from 'mobx'
+import { GameInfo } from 'common/types'
+import gameRequiresAccessCodes from 'frontend/helpers/gameRequiresAccessCodes'
+
+class UpdateState {
```
eliobricenov, line 5: Maybe change this to `GameUpdateState`? Update feels a bit too broad to me

eliobricenov (approved):
lgtm

src/frontend/state/UpdateState.ts
BrettCleary: yup. done

PR #856, opened 2024-04-10
[FIX] Search bar item click navigation

flavioislima:
Fix the current issue of failing to navigate to the game page from the search bar.

---

### DIFF
--- a/src/frontend/components/UI/SearchBar/index.tsx
+++ b/src/frontend/components/UI/SearchBar/index.tsx
@@ -85,11 +85,13 @@ export default observer(function SearchBar() {
 
       if (game !== undefined) {
         navigate(`/gamepage/${game.runner}/${game.app_name}`, {
-          state: { gameInfo: game }
+          state: { gameInfo: JSON.parse(JSON.stringify(game)) , fromDM: false}
         })
       }
     }
   }
+
+  const showAutoComplete = list.length > 0 && libraryState.filterText.length > 0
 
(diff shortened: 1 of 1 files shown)


eliobricenov (approved):

red-game-dev (approved):

BrettCleary (approved):

nyghtstalker:
1. There are no search results but the game is still showing

image

2. After clicking the search button the default selection is still the search icon instead of the field. Without the field selected the user has to select the first first before they start typing. Can we please have the field be selected after selecting the search icon so the user can start typing instead of having to select the field first?


image

flavioislima:
> 1. There are no search results but the game tile is still showing
> 
> image
> 
> 2. After clicking the search button the default selection is still the search icon instead of the field. Without the field selected the user has to select the first first before they start typing. Can we please have the field be selected after selecting the search icon so the user can start typing instead of having to select the field first?
> 
> 
> image
> 

1. The results are actually different from the auto complete. Because on the auto complete we only check the title exactly, but on the results on the library we use fuzzy search, so it will find the title that is closest to the search. So they don't need to match since they use different algorithms.

2. We will change the looks of this Search bar soon to be closer to the store so I won't change this on this PR.

nyghtstalker (approved):
Clicking the results now takes the user to the game page as expected. The other comments will be addressed later. Great job! I am excited to have the library search work now!

PR #857, opened 2024-04-11
[Feat] Add go back and go forward arrows to browser games

BrettCleary:
# Summary

Adds go back and go forward arrow icon buttons to browser games.

### Note
I attempted to set the disabled boolean based on `webview.canGoBack` but there was an issue attempting to use `webview.current` before the `"dom-ready"` event had fired. Correct me if I'm wrong, but I don't think this is a blocker.

image

# Testing
https://app.qase.io/project/HC?case=332&previewMode=modal

### DIFF
--- a/src/frontend/OverlayManager/index.module.scss
+++ b/src/frontend/OverlayManager/index.module.scss
@@ -1,9 +1,11 @@
+$topNavBarHeight: 30px;
+
 .browserGame {
   position: absolute;
   left: 0px;
   right: 0px;
   bottom: 0px;
-  top: 30px;
+  top: $topNavBarHeight;
 }
 
 .topBar {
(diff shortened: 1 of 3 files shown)


red-game-dev (approved):
LGTM 🙌

flavioislima:
> I attempted to set the disabled boolean based on webview.canGoBack but there was an issue attempting to use webview.current before the "dom-ready" event had fired. Correct me if I'm wrong, but I don't think this is a blocker.

i think we solved this on the main Webview.
One approach is to use a state for the disabled button and change it on a useEffect after the `dom-ready` event is received or something like it.

BrettCleary:
@flavioislima I completely forgot about that `WebviewControls` component. I'll just use that one instead!

image

Fixed the stacking issue with the hint text with the overlay

image

flavioislima (approved):
Looks great!

nyghtstalker (approved):
This is awesome! Super simple and works great! Awesome work!

PR #859, opened 2024-04-11
[UI/UX] Show store links on navbar on all pages and adjust it to small window sizes

flavioislima:
Small change to increase the discoverability of other stores on HyperPlay.
Also adjust the navbar and filters to small window sizes.

image

image

---

### DIFF
--- a/.github/workflows/flatpak-build.yml
+++ b/.github/workflows/flatpak-build.yml
@@ -52,6 +52,10 @@ jobs:
         run: yarn setup
       - name: Build artifacts.
         run: yarn dist:linux:ci:flatpak
+      - name: Clean up build files to save space
+        run: rm -rf build dist/linux-unpacked node_modules
+      - name: Install Axios
+        run: yarn add axios
       - name: Prepare Flatpak
         run: yarn flatpak:prepare
       - name: Build Flatpak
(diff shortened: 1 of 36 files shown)


eliobricenov (approved):

red-game-dev (approved):

flavioislima:
Ok so did some adjustments here to the minimum Window size:
Minimum supported resolution: Steam Deck one: 1280x800
Based on that, the minimum window size is equal ~66% or 1.5 of it. This way HP can still be opened with another APP and do not take the whole screen even on small screens like it. On 1080p screens resolution (most common ones) this will mean 44% of the screen. so seems like I reasonable size.

Based on that did the following changes for when the window is at the minimum size:
- Hide the HyperPlay Logo Text, kept only the logo itself
- Hide the Alpha badge
- Lowered the width of the Search bar when opened to 243px.

image

BrettCleary (approved):
nice. I think you need to run prettier again btw

flavioislima:
> nice. I think you need to run prettier again btw

this is happening more often than I want to. Maybe we should get back to add lint on the pre-push command.

nyghtstalker (approved):
This looks AWESOME! I love all these much needed improvements to a smaller screen! It was def needed when viewing the previous on Main! I love that the stores are now accessible from all the side nav pages! Great job!

image

PR #861, opened 2024-04-18
[TECH] Speed Up Big Libraries loading

flavioislima:
I've noticed that when I am logged in on all stores with more than 550 games the first loading takes almost a minute to show the games. This is due to how slow checking for game updates is since it iterates through all runners and wait for each list to complete to then full-fill the promise, so longer the library slower it will be.

So I did a few small changes:
- Added the a `Runner[]` parameter to `checkGameUpdates`.
- Await only for the list of HyperPlay games, so will only block the thread for these.
- Call it again for the rest of the runners but without blocking the thread.

This way the behaviour will be:
- The list of games will appear right away since the hyperplay request is quite fast.
- All other games from other stores will load but the update status for the games will only change after a few seconds.

This has other benefit as well since HP games will go up to the list for auto updates and since most people use it for our games then makes no sense to block everything because of other runners.

---

### DIFF
--- a/src/backend/api/misc.ts
+++ b/src/backend/api/misc.ts
@@ -46,8 +46,8 @@ export const getAmazonLoginData = async () =>
 export const authAmazon = async (data: NileRegisterData) =>
   ipcRenderer.invoke('authAmazon', data)
 export const logoutAmazon = async () => ipcRenderer.invoke('logoutAmazon')
-export const checkGameUpdates = async () =>
-  ipcRenderer.invoke('checkGameUpdates')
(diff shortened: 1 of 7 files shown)


eliobricenov (approved):
good stuff, I had to refresh my event loop knowledge here

BrettCleary (approved):
Good stuff!

red-game-dev (approved):

red-game-dev (commented):
Wonderful! 🙌 Well done 🍾

PR #874, opened 2024-04-30
[Chore] Keep app data on uninstall

BrettCleary:
Keep appdata on uninstall so extension data is not wiped out


flavioislima:
I'm only concern about this because if the user wants a complete uninstall of HyperPlay.
Because I think that the person will only uninstall the app if they either don't want it anymore or to try to fix some other issue.

Would be good to have an option on uninstall but I don't know if electron builder supports it.

red-game-dev:
> I'm only concern about this because if the user wants a complete uninstall of HyperPlay. Because I think that the person will only uninstall the app if they either don't want it anymore or to try to fix some other issue.
> 
> Would be good to have an option on uninstall but I don't know if electron builder supports it.

Just wondering on this one thing, there is 0 personal data in this that is being stored in any form right? Because of GDPR the following Articles

```
Article 5(1)(c) - Data Minimization: This article states that personal data shall be "adequate, relevant and limited to what is necessary in relation to the purposes for which they are processed.

Article 5(1)(e) - Storage Limitation: This mandates that personal data be kept in a form which permits identification of data subjects for no longer than is necessary for the purposes for which the personal data are processed.

Article 17 - Right to Erasure ('Right to be Forgotten'): This gives data subjects the right to have personal data erased without undue delay under certain circumstances, such as when the data is no longer necessary for the original purposes or when the data subject withdraws consent.
```

flavioislima:
> > I'm only concern about this because if the user wants a complete uninstall of HyperPlay. Because I think that the person will only uninstall the app if they either don't want it anymore or to try to fix some other issue.
> > 
> > Would be good to have an option on uninstall but I don't know if electron builder supports it.
> 
> Just wondering on this one thing, there is 0 personal data in this that is being stored in any form right? Because of GDPR the following Articles
> 
> ```
> Article 5(1)(c) - Data Minimization: This article states that personal data shall be "adequate, relevant and limited to what is necessary in relation to the purposes for which they are processed.
> 
> Article 5(1)(e) - Storage Limitation: This mandates that personal data be kept in a form which permits identification of data subjects for no longer than is necessary for the purposes for which the personal data are processed.
> 
> Article 17 - Right to Erasure ('Right to be Forgotten'): This gives data subjects the right to have personal data erased without undue delay under certain circumstances, such as when the data is no longer necessary for the original purposes or when the data subject withdraws consent.
> ```
> 
> 

That is not really the case as well.
The data on Appdata is related with the library of games, settings, etc. Not really user personal data since we don't collect it this way.

eliobricenov (approved):

flavioislima (approved):
ok, let's do it 👍🏽

red-game-dev (approved):

nyghtstalker:
Does one of these trigger a clear, fresh start of HP? If so, wouldn't this cause the same behavior as the uninstall with wiping appData? @flavioislima @BrettCleary 

image

PR #879, opened 2024-05-01
[Fix] Modal styling with dialog component

BrettCleary:
Fixes
- back to game and exit game buttons not clickable in overlay
- square outline on app launch as extension dialog element takes focus

Refactors the extension manager contents to a sub-component

PR #883, opened 2024-05-07
tech: use native UI for newsletter subscription

eliobricenov:

My first approach was to change the dev portal code since we're currently using a webview, but now that we're using Ortto and subscribing to the newsletter means only sending a POST request, I  believe it is best if we decouple this modal from the dev portal. The modal is in the UI library, we can use it in the app directly.

An old version of the launcher can still use the webview, keeping backward compatibility.

After we send the subscribe request, we set a flag in the store to prevent the modal from showing up again.

~**We need https://github.com/HyperPlay-Gaming/hyperplay-dev/pull/277 to be merged first**~

Steps to test: https://app.qase.io/case/HC-344

solves https://github.com/HyperPlay-Gaming/product-management/issues/327

---


flavioislima (approved):
Looks good

nyghtstalker:
1. This box appears after selecting my extension account

image

2. I accidently clicked away when the email modal appeared after selecting my extension account (the above image) and now I can't get that modal to trigger again... Do you only get ONE chance at this moment to enter your email to sign up for new??? If so, we def should add a way to trigger that again without needing to reset the extension again... 

3. I was able to get the correct modals to appear and told me I am now signed up. Is there any confirmation that this was successful that we can confirm? I don't seem to get anything via email after entering my email so that makes me think it didn't go through...?

eliobricenov:
> This box appeared again on launch after resetting extension for to confirm the modal does not appear again.

this is unrelated to these changes, the bug is in `main` and will be fixed in https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/pull/879 @nyghtstalker

eliobricenov:
>I accidently clicked away when the email modal appeared after selecting my extension account (the above image) and now I can't get that modal to trigger again... Do you only get ONE chance at this moment to enter your email to sign up for new??? If so, we def should add a way to trigger that again without needing to reset the extension again...

we decided we won't close the modal if the users click the backdrop, they need to click the close buttons

eliobricenov:
image

Confirmed that the request come through to Ortto

PR #888, opened 2024-05-15
[Fix] Fix sui support

BrettCleary:
Fixes sui support through the proxy server

# Testing 
Use these calls from https://docs.hyperplay.xyz/developer-docs/api-1/sui-blockchain/get-accounts while the app is running and metamask extension is connected


eliobricenov (approved):

red-game-dev (approved):

nyghtstalker (approved):
I can confirm that with the wallet connected the command triggers the request and starts the snap install.

PR #889, opened 2024-05-15
[Fix] Update for non token gated games

BrettCleary:
- If not token-gated, don't require a siwe signature on update
- Fix infinite update loop where the frontend library state was not being updated correctly on refresh
- Remove dead code from `cancelCurrentDownload` that was causing failure to cancel download in some cases


eliobricenov (approved):

red-game-dev (approved):

jiyuu-jin (approved):

nyghtstalker (approved):
Confirmed that you can now manually update with the icon and button.

PR #897, opened 2024-05-17
[FIX] Protocol Handler duplicated calls

flavioislima:
This PR should fix the intermittent issue of the Protocol handler being called twice from the store even though only one click event was emitted from the Play Game button.

## To test
- Use the artifact from the build, dev mode won't work.
- Install the app
- Go to the store and launch anything from there.
- Game should launch only once
- If not installed, the dialog asking to install should appear only once.

---


eliobricenov (commented):
this is the first time I see this part of the code, but maybe we don't need to execute the handleProtocol function? it looks like maybe we'd only want to focus the current window https://stackoverflow.com/a/52633637

flavioislima:
> this is the first time I see this part of the code, but maybe we don't need to execute the handleProtocol function? it looks like maybe we'd only want to focus the current window https://stackoverflow.com/a/52633637
> 

I tried it and it doesn't work.

The problem is not the second instance but the fact that the second-instance event is being triggered twice. So it calls the method two times as well.

flavioislima:
> I'd say this is a good temp fix but I think we should spend more time investigating why the command is executing twice (I also hate bugs like this 😭)

This one is tricky because it's not happening all the time as well.

flavioislima:
Now when using the client the Store will also use the `window.api` to launch or install the game, so this will make it more predictable and fix some edge cases.

BrettCleary:
Note that we also need to test this PR with the store URLs pointing to store.hyperplay.xyz since it should be merged prior to the store PR

BrettCleary (approved):
install and play is working for me for browser and native games.

left some UX comments on the store PR 

also reminder to revert back to store.hyperplay.xyz before merging and that we need to merge this PR and release before merging the store PR

flavioislima:
> Note that we also need to test this PR with the store URLs pointing to store.hyperplay.xyz since it should be merged prior to the store PR

yeah, good point, otherwise merging the store first will break production.

red-game-dev (approved):

red-game-dev:
Amazing job @flavioislima, cheers for finding a way for this. It was dam tricky 🙌

flavioislima:
> Amazing job @flavioislima, cheers for finding a way for this. It was dam tricky 🙌 

Yeah, actually using the window.api was Brett's idea and it works a lot better.

Based on that I'm planning on implementing other methods to improve the communication between the store and the client in the future.

nyghtstalker:
Play button on the store page remains inaccessible after install is complete through the store page.

image

image

flavioislima:
> Play button on the store page remains inaccessible after install is complete through the store page.
> 

Yes, I am working on new changes to improve this communication between the store and the client because it doesn't know if the game was installed or not until you get out of the page so we need to implement an IPC call for that but I will do it on another PR.

nyghtstalker (approved):
Button not working will be addressed in another PR. This looks good!

PR #898, opened 2024-05-18
[Tech] Refactor extension helper

BrettCleary:
Refactor extension helper

## Testing
- https://app.qase.io/project/HC?suite=48&case=327&previewMode=modal
- Test import, create new wallet, create wallet with recovery phrase, and manual import with onboarding modal
- Test reset extension in settings on all platforms
- https://app.qase.io/project/HC?suite=48&previewMode=modal&case=261


buildResetExtension.js
```diff
@@ -0,0 +1,28 @@
...
+const { existsSync } = require('fs')
+const { platform } = require('os')
+const process = require('process')
+
+const [nodeExe, scriptPath, ...args] = process.argv
+
+const resetExtensionScript =
+  './node_modules/@hyperplay/extension-importer/src/resetExtension.js'
+if (existsSync(resetExtensionScript)) {
+  console.log('spawning pkg', resetExtensionScript, args)
```
flavioislima, line 11: Should we keep this?

flavioislima (approved):
Codewise looks good. Need to test a lot to be sure everything is working on all platforms, overlay and on builds as well.

buildResetExtension.js
BrettCleary: It's only logged in the build step in the github action so I think it could be useful for debugging

red-game-dev (approved):

nyghtstalker:
1. Clicking "Forgot password?" maximizes the client. These options in the wallet modal open in maximized. Expand View and Support open in maximized. Lots of different things open in maximized.

image
image
image

2. Changing the network hides the option after

image
image

3. Hovering over the address triggers scroll bars to appear and cuts of text

image

4. Wallet address cuts past container

image

BrettCleary:
@nyghtstalker 1 is addressed now. 

I can't replicate 2 but the reason you're seeing it is because the network dropdown is shown on the popup lock screen but not the notification page lock screen. This is intended behavior on the MetaMask side

 3 and 4 are fine for now/out of scope

BrettCleary:
@nyghtstalker 4 is actually the same problem on the https://portfolio.metamask.io/ site too. so metamask needs to fix

nyghtstalker:
On Windows - when using the latest artifact the "Reset Extension" option does not work. It does not reset it or relaunch the client.

nyghtstalker (approved):

BrettCleary:
going to disable import srp direct link button on separate PR

PR #902, opened 2024-05-20
[Fix] Get connected provider

BrettCleary:
Fix provider import causing MM extension to not display in browser games


flavioislima (approved):
Good stuff!

nyghtstalker (approved):
Confirmed I am now seeing MM extension when launching browser games.

image

PR #905, opened 2024-05-22
[Fix] browser game mm links

BrettCleary:
This one line fix for browser game metamask popup links turned into a larger change since the constants file import became a circular dependency.

To fix, the hpApi constants imports were moved into a separate folder.


flavioislima (approved):
lgtm

PR #911, opened 2024-05-25
[UI] Update download manager page

BrettCleary:
Update download manager to latest designs

### DIFF
--- a/src/frontend/screens/DownloadManager/index.tsx
+++ b/src/frontend/screens/DownloadManager/index.tsx
@@ -6,9 +6,9 @@ import { DMQueueElement, DownloadManagerState } from 'common/types'
 import { UpdateComponent } from 'frontend/components/UI'
 import ProgressHeader from './components/ProgressHeader'
 import DownloadManagerHeader from './DownloadManagerHeader'
-import { downloadManagerStore } from 'frontend/helpers/electronStores'
 import { DMQueue } from 'frontend/types'
 import DownloadManagerItem from './components/DownloadManagerItem'
+import { Tabs, getTabsClassNames } from '@hyperplay/ui'
 
 export default React.memo(function DownloadManager(): JSX.Element | null {
   const { t } = useTranslation()
@@ -63,11 +63,6 @@ export default React.memo(function DownloadManager(): JSX.Element | null {
     return <UpdateComponent />
   }
 
-  const handleClearList = () => {
-    setFinishedElem([])
-    downloadManagerStore.set('finished', [])
-  }
-
   const doneElements =
     (finishedElem?.length &&
       finishedElem.sort((a, b) => {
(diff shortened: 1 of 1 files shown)


flavioislima:
Testing here.
Overall pretty nice changes.

One thing I noticed on the Downloaded tab is that there is no action button. Perhaps adding a Play button there for the finished would be nice. It could either launch the game or navigate to the game page that is simpler to implement.

image

flavioislima:
Another thing I noticed is that the Cancel button doesn't show the cancel dialog anymore. It simply removes the game from the list without confirmation.

eliobricenov (approved):

src/frontend/screens/DownloadManager/index.tsx
```diff
@@ -84,81 +77,61 @@ export default React.memo(function DownloadManager(): JSX.Element | null {
...
-            </span>
-            <div className="dmItemList">
-              <DownloadManagerHeader time="finished" />
-              {doneElements.map((el, key) => (
-                <DownloadManagerItem key={key} element={el} current={false} />
-              ))}
+          </Tabs.Panel>
+
+          <Tabs.Panel value="queued">
+            <>
```
eliobricenov, line 119: we can remove the `<></>` since there is only one child node

src/frontend/screens/DownloadManager/index.tsx
```diff
@@ -84,81 +77,61 @@ export default React.memo(function DownloadManager(): JSX.Element | null {
...
-          {currentElement && (
-            <div className="downloadManager">
+    <>
+      <Background style={{ position: 'absolute' }}></Background>
+      <div className="contentContainer">
+        <h3
+          style={{
+            textAlign: 'left'
+          }}
+        >
```
eliobricenov, line 87: do we need this to be inline?

eliobricenov (approved):
code lgtm, I only have a couple comments

BrettCleary:
> Testing here. Overall pretty nice changes.
> 
> One thing I noticed on the Downloaded tab is that there is no action button. Perhaps adding a Play button there for the finished would be nice. It could either launch the game or navigate to the game page that is simpler to implement.
> 
> image

ah. I forgot to pass some state to the download manager table and to set the fill on the icons. updated now. icons centered and download icon switched over as well

BrettCleary:
> Another thing I noticed is that the Cancel button doesn't show the cancel dialog anymore. It simply removes the game from the list without confirmation.

can you try again? I cannot replicate this with the latest changes
image

src/frontend/screens/DownloadManager/index.tsx
BrettCleary: nope. I will move it to a class

src/frontend/screens/DownloadManager/index.tsx
BrettCleary: nice catch

BrettCleary:
I believe all comments are addressed @flavioislima @eliobricenov

eliobricenov (approved):
lgtm :)

red-game-dev (approved):

flavioislima (approved):
Nice! looks perfect now!

nyghtstalker:
1. "Nothing to download" is no longer displayed like before

image

2. Download graph remains after download finishes and after canceling the download. It is displayed on all tabs.

image
image
image

3. Clicking the game name or Play button takes the user to the game page. Using the back button or mouse back button takes the user to the Downloading page instead of the Downloaded page they were on previously.

image
image

4. Cancels are not displayed in the list

image

BrettCleary:
> 1. "Nothing to download" is no longer displayed like before
> 
> 2. Download graph remains after download finishes and after canceling the download. It is displayed on all tabs.
> 
> 
> 3. Clicking the game name or Play button takes the user to the game page. Using the back button or mouse back button takes the user to the Downloading page instead of the Downloaded page they were on previously.
> 
> 
> 4. Cancels are not displayed in the list
> 

1. this is fine. showed @pimentel08 and got approval for this iteration as well
2. I think this is fine too. probably better than having layout shift showing and hiding it, but we can send a follow up pr optimizing that if needed
3. I think this is fine too and a small optimization that will take more engr time than it is worth
4. I can see cancels on mine

image

nyghtstalker (approved):
Ah! Canceling during extraction looks to be the issue. If you cancel during extraction it shows up as successful and not canceled. I first canceled during download and it did show up as canceled in the list. Then I canceled it during extraction and the canceled went away. Unfortunately, this same behavior is in PROD.

PR #912, opened 2024-05-29
[Feat] Quests overlay reward claim

BrettCleary:
# Summary
This one got a lil big. ~600 lines. It is mostly feature flagged new features though

This PR
- Adds quest reward minting in the overlay with wagmi
- Adds wagmi to index.tsx so it can be used in the future in the main window as well
- Upgrades to `@tanstack/react-query` which is the new package name for react-query v4
- Adds support for HP account sign in through the overlay
- Refactors `QuestViewer` into `QuestDetailsWrapper` and `QuestLogWrapper` since it was getting bloated after adding the minting code

# Note
The UX is not perfect and that is not the goal of this PR. There will be follow up PR's for 
- getting claimable/claimed status on initial page load
- overlay hiding after confirming switch chain/approve txn
- adding support for more chains in wagmi

# Testing
Test with the overlay in Kosium


flavioislima (approved):
Good stuff

src/backend/api/auth.ts
```diff
@@ -12,3 +12,15 @@ export const logOut = async () => {
 
 export const signInWithProvider = (provider: string) =>
   ipcRenderer.send('openExternalUrl', `${DEV_PORTAL_URL}/oauth/${provider}`)
+
+export const authConnected = () => ipcRenderer.send('authConnected')
+export const authDisconnected = () => ipcRenderer.send('authDisconnected')
```
eliobricenov, line 17: What do you think about renaming these to `notifyWindowsAuthConnected` and `notifyWindowsauthDisconnected`?

This is what we're doing right?

eliobricenov (approved):
lgtm, great work

I left a couple suggestions that can come up in a follow up PR

red-game-dev (approved):
LGTM - Checked only the code

nyghtstalker:
1. All font is larger

nyghtstalker:
1. All font is larger

image
image

2. Attempting to claim with the wallet locked has no effect. I would recommend some way of alerting the user to unlock their wallet.

------------------

Other than the above I was able to claim quests and sign into my email from the overlay.

image

nyghtstalker (approved):

src/backend/api/auth.ts
BrettCleary: it is notifying the backend only. the backend decides whether to notify other windows or not

BrettCleary:
> lgtm, great work
> 
> I left a couple suggestions that can come up in a follow up PR

I will definitely handle these on a follow-up PR 🙏

PR #914, opened 2024-05-31
[Feat] Navigate to Game page from Store

flavioislima:
Store PR: https://github.com/HyperPlay-Gaming/hyperplay-store/pull/136
Needs to be released at the same time as the Store PR.

Adds Api to give support to navigate to the library game page from the store.

---

### DIFF
--- a/src/frontend/components/UI/WebviewControls/index.tsx
+++ b/src/frontend/components/UI/WebviewControls/index.tsx
@@ -85,7 +85,7 @@ export default function WebviewControls({
   )
 
   const _url = url !== '' ? new URL(url) : null
-  const allowList = ['store.hyperplay.xyz', 'docs.hyperplay.xyz']
+  const allowList = ['hyperplay-store-git-feat-libfromstore-hyperplay.vercel.app', 'docs.hyperplay.xyz']
 
   if (_url && allowList.includes(_url.host)) return null
 
--- a/src/frontend/components/UI/WebviewControls/index.tsx
+++ b/src/frontend/components/UI/WebviewControls/index.tsx
@@ -85,7 +85,7 @@ export default function WebviewControls({
   )
 
   const _url = url !== '' ? new URL(url) : null
-  const allowList = ['hyperplay-store-git-feat-libfromstore-hyperplay.vercel.app', 'docs.hyperplay.xyz']
+  const allowList = ['store.hyperplay.xyz', 'docs.hyperplay.xyz']
 
   if (_url && allowList.includes(_url.host)) return null
 
--- a/src/frontend/constants.ts
+++ b/src/frontend/constants.ts
@@ -1,6 +1,5 @@
 export const EPIC_LOGIN_URL = 'https://legendary.gl/epiclogin'
-export const HYPERPLAY_STORE_URL =
-  'https://hyperplay-store-git-feat-libfromstore-hyperplay.vercel.app?isLauncher=true'
+export const HYPERPLAY_STORE_URL = 'https://store.hyperplay.xyz?isLauncher=true'
 export const EPIC_STORE_URL = 'https://www.epicgames.com/store'
 export const GOG_STORE_URL = `https://gog.com`
 export const WIKI_URL = 'https://docs.hyperplay.xyz/'
--- a/src/frontend/constants.ts
+++ b/src/frontend/constants.ts
@@ -1,5 +1,6 @@
 export const EPIC_LOGIN_URL = 'https://legendary.gl/epiclogin'
-export const HYPERPLAY_STORE_URL = 'https://store.hyperplay.xyz?isLauncher=true'
+export const HYPERPLAY_STORE_URL =
+  'https://hyperplay-store-git-feat-libfromstore-hyperplay.vercel.app?isLauncher=true'
 export const EPIC_STORE_URL = 'https://www.epicgames.com/store'
 export const GOG_STORE_URL = `https://gog.com`
 export const WIKI_URL = 'https://docs.hyperplay.xyz/'
--- a/src/frontend/screens/WebView/index.tsx
+++ b/src/frontend/screens/WebView/index.tsx
@@ -32,7 +32,7 @@ import { getGameInfo } from 'frontend/helpers'
 
 function urlIsHpUrl(url: string) {
   const urlToTest = new URL(url)
-  return urlToTest.hostname === 'store.hyperplay.xyz'
+  return urlToTest.hostname === 'hyperplay-store-git-feat-libfromstore-hyperplay.vercel.app'
 }
 
 function shouldInjectProvider(url: string) {
--- a/src/frontend/screens/WebView/index.tsx
+++ b/src/frontend/screens/WebView/index.tsx
@@ -126,6 +126,10 @@ function WebView() {
   }, [startUrl, runner])
 
   useEffect(() => {
+    if (!urlIsHpUrl(startUrl)) {
+      return 
+    }
+
     window.api.handleGoToGamePage(async (_, gameId) => {
       const gameInfo = await getGameInfo(gameId, 'hyperplay')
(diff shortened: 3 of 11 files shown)


src/frontend/constants.ts
```diff
@@ -1,5 +1,6 @@
 export const EPIC_LOGIN_URL = 'https://legendary.gl/epiclogin'
-export const HYPERPLAY_STORE_URL = 'https://store.hyperplay.xyz?isLauncher=true'
+export const HYPERPLAY_STORE_URL =
+  'https://hyperplay-store-git-feat-libfromstore-hyperplay.vercel.app?isLauncher=true'
```
eliobricenov, line 3: we need https://github.com/HyperPlay-Gaming/hyperplay-store/pull/136 to be merged

src/frontend/screens/WebView/index.tsx
```diff
@@ -28,10 +28,11 @@ import {
 } from '../../constants'
 import { METAMASK_SNAPS_URL } from 'common/constants'
 import storeAuthState from 'frontend/state/storeAuthState'
+import { getGameInfo } from 'frontend/helpers'
 
 function urlIsHpUrl(url: string) {
   const urlToTest = new URL(url)
-  return urlToTest.hostname === 'store.hyperplay.xyz'
+  return urlToTest.hostname === 'hyperplay-store-git-feat-libfromstore-hyperplay.vercel.app'
```
eliobricenov, line 35: same here

src/frontend/screens/WebView/index.tsx
```diff
@@ -124,6 +125,14 @@ function WebView() {
     window.api.trackScreen('WebView', { url: startUrl, runner })
   }, [startUrl, runner])
 
+  useEffect(() => {
+    window.api.handleGoToGamePage(async (_, gameId) => {
+      const gameInfo = await getGameInfo(gameId, 'hyperplay')
+      navigate(`/gamepage/hyperplay/${gameId}`, { state: { gameInfo, fromDM: false } })
+    }
+    )
+  }, [])
```
eliobricenov, line 134: I don't understand this change 🤔

Why do we need to navigate to the game page when on mount? This component is used for multiple webviews

```tsx
<Route
    path="hyperplaystore"
    element={<WebView key="hyperplaystore" />}
/>
<Route path="epicstore" element={<WebView key="epicstore" />} />
<Route path="gogstore" element={<WebView key="gogstore" />} />
<Route path="wiki" element={<WebView key="wiki" />} />
<Route path="metamaskHome" element={<MetaMaskHome />} />
<Route
    path="metamaskSnaps"
    element={<WebView key="metamaskSnaps" />}
/>
```

src/frontend/components/UI/WebviewControls/index.tsx
```diff
@@ -85,7 +85,7 @@ export default function WebviewControls({
   )
 
   const _url = url !== '' ? new URL(url) : null
-  const allowList = ['store.hyperplay.xyz', 'docs.hyperplay.xyz']
+  const allowList = ['hyperplay-store-git-feat-libfromstore-hyperplay.vercel.app', 'docs.hyperplay.xyz']
```
eliobricenov, line 88: same here

eliobricenov (changes requested):

src/frontend/constants.ts
flavioislima: yeah, added this info to the other PR as well. thanks

flavioislima: This is actually just for testing, I will revert these changes

src/frontend/screens/WebView/index.tsx
flavioislima: Not really, this just adds a listener on mount.
Otherwise when we call this method from the store it won't work.

So basically when we mount the Webview we start listening to these calls.

But that is true that will be for all URLs, I will add a check to only add the listener for hyperplay store.

src/frontend/screens/WebView/index.tsx
```diff
@@ -124,6 +125,18 @@ function WebView() {
...
+    if (!urlIsHpUrl(startUrl)) {
+      return 
+    }
+
+    window.api.handleGoToGamePage(async (_, gameId) => {
+      const gameInfo = await getGameInfo(gameId, 'hyperplay')
+      navigate(`/gamepage/hyperplay/${gameId}`, { state: { gameInfo, fromDM: false } })
+    }
+    )
+  }, [])
```
BrettCleary, line 138:
```suggestion
    const removeHandleGoToGamePage = window.api.handleGoToGamePage(async (_, gameId) => {
      const gameInfo = await getGameInfo(gameId, 'hyperplay')
      navigate(`/gamepage/hyperplay/${gameId}`, { state: { gameInfo, fromDM: false } })
    }
    )
    
    //useEffect unmount
    return () => {
      removeHandleGoToGamePage()
    }
  }, [])
```

need to unmount or else it will add another listener on every render of the webview

BrettCleary (approved):
it works! 🥳 

couple comments:
- you need to remove the listener on the useEffect unmount hook on this PR
- we need to update the game page to the latest designs on a separate PR
- it would be nice if the install modal were shown after navigating to the game page in the library. to save the user a click.

src/frontend/screens/WebView/index.tsx
flavioislima: Good catch, We do that on the api side so I thought we didn't need it here as well just like we are doing with the handleGameStatus, but makes sense to have it here as well.

src/frontend/constants.ts
flavioislima: reverted

src/frontend/screens/WebView/index.tsx
flavioislima: done

flavioislima:
> * it would be nice if the install modal were shown after navigating to the game page in the library. to save the user a click.

I will open another PR after the next release for this.

src/frontend/screens/WebView/index.tsx
```diff
@@ -124,6 +125,32 @@ function WebView() {
...
+    })
+
+    const removeHandleGoToGamePage = window.api.handleGoToGamePage(
+      async (_, gameId) => {
+        const gameInfo = await getGameInfo(gameId, 'hyperplay')
+        navigate(`/gamepage/hyperplay/${gameId}`, {
+          state: { gameInfo, fromDM: false }
+        })
+      }
+    )
```
BrettCleary, line 147: this is duplicated now

BrettCleary (changes requested):

src/frontend/screens/WebView/index.tsx
flavioislima: yeah, i am really wondering if we need to remove the call here if we are doing that on the api:
```export const handleGoToGamePage = (
  onChange: (e: Electron.IpcRendererEvent, gameId: string) => void
) => {
  ipcRenderer.on('goToGamePage', onChange)
  return () => {
    ipcRenderer.removeListener('goToGamePage', onChange)
  }
}```

BrettCleary: we are not unsubscribing in the api fxn definition. we are retruning a callback that we can call later on to unsubscribe the listener

flavioislima: So just retuning the same method should unsubscribe, no? we need a different method for that?

flavioislima: Got it, pushed a fix.

BrettCleary (approved):

eliobricenov (approved):
LGTM

reminder to change the URL (I merged a PR with the preview URL 😅)

red-game-dev (approved):
LGTM to me, checked the code only though 🙏

nyghtstalker (approved):
This is awesome! Clicking the "Add to Library" and "Play" buttons on the store game pages now takes the user to the library game page and no longer launches the game from the store. The user must now launch the game only from the library.

IvanVnn (approved):

PR #926, opened 2024-06-14
[Fix] Compatibility with deprecated on topics and fix send request return for browser game provider

BrettCleary:
# Summary
Add deprecated provider on topics for backwards compatibility. Return result directly from send

# Testing
https://app.qase.io/project/HC?suite=61&case=346&previewMode=modal

### DIFF
--- a/src/backend/proxy/providerPreload.ts
+++ b/src/backend/proxy/providerPreload.ts
@@ -17,7 +17,10 @@ const enabledTopics = [
   'connect',
   'message',
   'disconnect',
-  'chainChanged'
+  'chainChanged',
+  'close',
+  // deprecated
+  'networkChanged'
 ]
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
@@ -53,7 +56,7 @@ const sendRequest = async (...args: unknown[]) => {
   if (Object.hasOwn(result, 'error')) {
     throw result[0].error.message
   }
-  return result[0].result
+  return result
 }
 
 const sendAsyncRequest = async (payload: any, callback: JsonRpcCallback) => {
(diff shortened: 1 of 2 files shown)


red-game-dev (approved):

jiyuu-jin (approved):

src/backend/proxy/providerPreload.ts
```diff
@@ -53,7 +56,7 @@ const sendRequest = async (...args: unknown[]) => {
   if (Object.hasOwn(result, 'error')) {
     throw result[0].error.message
   }
-  return result[0].result
+  return result
```
flavioislima, line 59: Is this related? Instead of a value how it will return an array, right?
Where do we declare the types for it?

flavioislima (approved):
Looks good, left a small question though.

src/backend/proxy/providerPreload.ts
BrettCleary: The return type is `unknown` https://eips.ethereum.org/EIPS/eip-1193#send-deprecated

`Provider.send(...args: unknown[]): unknown;`

I'm not sure why I was treating it as an array before. I believe we may have been batching these at some point but aren't any more and it was a regression? Either way, we need to return the exact value we receive from the provider which is what result is currently.

I don't think lines 56-58 are needed either. I can remove them

BrettCleary: also from testing, this method actually throws now with the same error instead of returning an error prop. I believe this change was made a while back but this method was never updated

BrettCleary: and yes it is related because it is breaking one of our browser games

nyghtstalker (approved):
Confirmed I am able to connect the wallet to DeFi via the overlay. It works great!

PR #927, opened 2024-06-25
[Feat] Add reward claiming for other reward types

BrettCleary:
# Summary
- points claim
- external task complete
- refactor to separate modules for each reward claim type
- use new endpoint for getting user playstreak
- add `fetchWithCookie` for code reuse
- add LD flag for overlay claim CTA
- show current streak = required streak when current > required
- add `gamesToShowQuestsFor` flag
- fix wagmi chain urls, add polygon
- add erc721 and erc1155 minting (will need backend changes before 721 can be tested e2e)

# Testing
- test points claim with https://app.qase.io/project/HC?case=351
- 20, 721, 1155, and external task cannot be tested yet


src/backend/utils/fetch_with_cookie.ts
```diff
@@ -0,0 +1,18 @@
+import { DEV_PORTAL_URL } from 'common/constants'
+import getPartitionCookies from './get_partition_cookies'
+
+export async function fetchWithCookie(url: string, method: string) {
+  const cookieString = await getPartitionCookies({
+    partition: 'persist:auth',
```
eliobricenov, line 6: Can we add the partition as the argument?

src/backend/utils/fetch_with_cookie.ts
```diff
@@ -0,0 +1,18 @@
...
+    url: DEV_PORTAL_URL
+  })
```
eliobricenov, line 8: we should probably return the fetch to handle `result.ok` in the implementations

```ts

const response = await fetchWithCookie(url, 'POST')

if (!response.ok) {
  ...
}

```

eliobricenov (commented):
lgtm, I left two comments for the fetch helper

src/backend/utils/fetch_with_cookie.ts
BrettCleary: yup. good idea

src/backend/utils/fetch_with_cookie.ts
BrettCleary: hmm this could get pretty repetitive though. what do you think about adding this in the fetchWithCookie function

eliobricenov (approved):

red-game-dev (approved):

nyghtstalker (approved):
I am able to claim the streak, the streak resets after claiming, and I am unable to start the streak again. Great job!

PR #932, opened 2024-07-09
[Tech] Add window.api.otp in dev and debug mode

BrettCleary:
# Summary

allow for `window.api.otp` to be called in debug and dev mode. Otherwise testing auth deeplinks will only work in the packaged app, which takes time to build and slows down development.


eliobricenov (approved):

nyghtstalker (approved):
I was able to use the deep links in dev mode. Looks good!

image

PR #940, opened 2024-07-16
[Tech] get test quests in dev/debug modes

BrettCleary:
# Summary
- show test quests in dev and debug modes


nyghtstalker (approved):
I am able to see the new quests and trigger the playstreak.

image

PR #942, opened 2024-07-16
[Tech] Fix channel logs

BrettCleary:
# Summary
- change `console.error` to `logError` so that error message can be seen in production for channel names that were changed or removed
- early return so that library refresh does not get stuck if a library game is in this state

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hyperplay",
-  "version": "0.16.0",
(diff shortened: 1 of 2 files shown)


flavioislima (approved):
lgtm

eliobricenov (approved):

red-game-dev (approved):

PR #943, opened 2024-07-16
[Fix] go to game page instead of play

BrettCleary:
# Summary
- go to game page on quest details button click instead of window.api.launch since that doesn't handle all the cases like if a game is already running or is uninstalled


eliobricenov (approved):

red-game-dev (approved):

src/frontend/screens/Quests/components/QuestDetailsViewPlay/index.tsx
```diff
@@ -146,13 +146,13 @@ export function QuestDetailsViewPlayWrapper({
...
       classNames={{ root: styles.questDetailsRoot }}
       isQuestsPage={true}
-      onPlayClick={async () =>
-        window.api.launch({
-          appName: questMeta.project_id,
-          launchArguments: '',
-          runner: 'hyperplay'
+      onPlayClick={async () => {
+        const gameId = questMeta.project_id
+        const gameInfo = await getGameInfo(gameId, 'hyperplay')
```
flavioislima, line 151: I think this is fine as a workaround for now but for some people this will cause a few seconds delay, maybe more.
Ideally we should be able to get the gameinfo when the component mount and then use it from there because would make the navigation instantly.

flavioislima (approved):
Works fine just left a small comment.

nyghtstalker (approved):
Play button has been removed from the quests page. The user now needs to go to the game page first to launch.

image

PR #949, opened 2024-07-22
feat: add all chains from @hyperplay/chains to wagmi config

eliobricenov:

This is a copy/paste of the code we have in the dev portal.

---

### DIFF
--- a/src/frontend/config.ts
+++ b/src/frontend/config.ts
@@ -1,77 +1,6 @@
 import { createConfig, http } from 'wagmi'
 import { Chain, hardhat, mainnet, polygon } from 'wagmi/chains'
-import { chainMap } from '@hyperplay/chains'
-
-type RpcUrls = {
-  http: string[]
-  webSocket?: string[]
-}
-
-type ChainMetadata = typeof chainMap[keyof typeof chainMap]
-
-export function parseChainMetadataToWagmiChain(
-  chainMetadata: ChainMetadata
-): Chain {
-  const validRpcs = chainMetadata.chain.rpc.filter(
-    (rpc) => !rpc.includes(`\${`)
-  )
-
-  const defaultRpcUrl = validRpcs[0]
-
-  if (!defaultRpcUrl) {
-    throw new Error('defaultRpcUrl is not defined')
-  }
-
-  const rpcUrls: Record<string, RpcUrls> = validRpcs.reduce((acc, url) => {
-    acc[url] = {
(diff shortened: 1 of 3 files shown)


src/frontend/config.ts
```diff
@@ -1,11 +1,101 @@
...
+      }, {} as Record<string, { name: string; url: string }>)
+
+    wagmiChain.blockExplorers = {
+      default: defaultExplorer,
+      ...otherExplorers
+    }
+  }
+
+  return wagmiChain
+}
```
eliobricenov, line 74: I tried to use the function we have in the chains package but I was getting this error. Probably due to a mismatch int the viem versions.

I will address it in a follow up PR.
image

src/frontend/config.ts
```diff
@@ -1,11 +1,101 @@
...
+import { chainMap } from '@hyperplay/chains'
+
+type RpcUrls = {
+  http: string[]
+  webSocket?: string[]
+}
+
+type ChainMetadata = typeof chainMap[keyof typeof chainMap]
+
+export function parseChainMetadataToWagmiChain(
```
BrettCleary, line 12: shouldn't we move this to hp chains and export from there. making `wagmi/chains` a dev dependency?

src/frontend/config.ts
```diff
@@ -1,11 +1,101 @@
...
+    transports[chain.id] = http(chainMetadata.chain.rpc[0])
+    // eslint-disable-next-line no-empty
+  } catch (error) {}
+}
+
+// add hardhat chain for development
+chainsToSupport = chainsToSupport.filter((chain) => chain.id !== hardhat.id)
+chainsToSupport.push(hardhat)
 
 export const config = createConfig({
```
BrettCleary, line 93: maybe we should just export this from hp chains too for code reuse. unless we think it might change independently of the dev portal

src/frontend/config.ts
BrettCleary: Why not fix it in this PR? We can make viem a shared dependency in hp chains too if it is breaking stuff like this

src/frontend/config.ts
eliobricenov: I think it is fine to leave these here for now

BrettCleary (approved):

jiyuu-jin:
LGTM 🚀

jiyuu-jin (approved):

PR #952, opened 2024-07-23
[Fix] Stop game multiple processes

BrettCleary:


flavioislima:
This should fix the edge case of some games that uses a launcher or a bootstrap before launching the actual game executable and then when the process ends or gets aborted it won't close all child processes: example: XOciety.

src/backend/launcher.ts
```diff
@@ -964,6 +986,77 @@ async function callRunner(
...
+  if (isWindows) {
+    logWarning(
+      `Killing all processes spawned by PID ${childPid}`,
+      LogPrefix.Backend
+    )
+
+    try {
+      // Get the list of child processes
+      const getChildProcesses = async (parentPid: number) => {
+        const result = await spawnAsync('powershell.exe', [
```
jiyuu-jin, line 999: Will this only work on windows?

jiyuu-jin (approved):

src/backend/launcher.ts
flavioislima: This part yes, at the end of the method you will find the command to mac and linux.

flavioislima (approved):

flavioislima:
@nyghtstalker the best game to test this one is Xociety.
Launch the game on the stable version of hyperplay and click the Exit button on the overlay after a few seconds, the game will keep opened.

Then try the same on this PR, it will take up to 5 seconds but will close it just fine.

PR #953, opened 2024-07-24
feat(quest):  add balance check and claim error CTA message 

eliobricenov:
## Summary

* Add a warning message when the connected wallet doesn't have enough gas to cover the transaction

Screenshot 2024-07-24 at 4 07 57 PM

* Add an error message when there is an error claiming the quest

Screenshot 2024-07-24 at 4 15 07 PM

### How to test

* Balance check: https://app.qase.io/case/HC-352
* Claim error: https://app.qase.io/case/HC-353

---

### DIFF
--- a/src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
+++ b/src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { useEffect, useState } from 'react'
 import {
   QuestDetails,
   QuestDetailsProps,
@@ -95,7 +95,8 @@ export function QuestDetailsWrapper({
   const {
     writeContract,
     error: writeContractError,
-    isPending: isPendingWriteContract
+    isPending: isPendingWriteContract,
+    reset: resetWriteContract
   } = useWriteContract()
 
   const {
@@ -311,6 +312,11 @@ export function QuestDetailsWrapper({
     claimRewardsMutation.isPending ||
     isPendingWriteContract ||
     isPendingSwitchingChain
+
+  useEffect(() => {
+    setWarningMessage(undefined)
+    resetWriteContract()
+  }, [selectedQuestId])
 
   if (selectedQuestId !== null && questMeta !== undefined) {
     const ctaDisabled =
--- a/src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
+++ b/src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
@@ -10,7 +10,13 @@ import useGetQuest from 'frontend/hooks/useGetQuest'
 import useGetSteamGame from 'frontend/hooks/useGetSteamGame'
 import useAuthSession from 'frontend/hooks/useAuthSession'
 import { useTranslation } from 'react-i18next'
-import { useWriteContract, useAccount, useSwitchChain } from 'wagmi'
+import {
+  useWriteContract,
+  useAccount,
+  useSwitchChain,
+  http,
+  useBalance
+} from 'wagmi'
 import { Reward } from 'common/types'
 import authState from 'frontend/state/authState'
 import { mintReward } from './rewards/mintReward'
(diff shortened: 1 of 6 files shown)


src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
```diff
@@ -141,20 +206,45 @@ export function QuestDetailsWrapper({
...
+
+    const gasNeeded = await getRewardClaimGasEstimation(reward)
+    const hasEnoughBalance = walletBalance.value >= gasNeeded
+
+    if (!hasEnoughBalance) {
+      window.api.logError(
+        `Not enough balance in the connected wallet to cover the gas fee associated with this Quest Reward claim. Current balance: ${walletBalance.value}, gas needed: ${gasNeeded}`
+      )
+      setWarningMessage(
+        'Not enough balance in the connected wallet to cover the gas fee associated with this Quest Reward claim.'
```
BrettCleary, line 238: should we translate this string? not sure if gas fee will translate correctly

src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
```diff
@@ -25,23 +31,82 @@ import { getRewardCategory } from 'frontend/helpers/getRewardCategory'
...
 import { getPlaystreakArgsFromQuestData } from 'frontend/helpers/getPlaystreakArgsFromQuestData'
+import { createPublicClient } from 'viem'
+import { chainMap, parseChainMetadataToViemChain } from '@hyperplay/chains'
+import { InfoAlertProps } from '@hyperplay/ui/dist/components/AlertCard'
 
 export interface QuestDetailsWrapperProps {
   selectedQuestId: number | null
 }
 
+const averageGasPerFunction: Record<string, number> = {
```
BrettCleary, line 42: it isn't the average gas per function though right? it's the estimated gas per function here. maybe we should rename

src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
```diff
@@ -25,23 +31,82 @@ import { getRewardCategory } from 'frontend/helpers/getRewardCategory'
...
+    transport: http()
+  })
+
+  let gasPerFunction
+
+  switch (reward.reward_type) {
+    case 'ERC1155':
+    case 'ERC721':
+    case 'ERC20':
+      gasPerFunction = averageGasPerFunction[reward.reward_type]
```
BrettCleary, line 73: what do you think about just increasing this by 50% to be on the safe side so when we make contract iterations in the future, this code will still probably work. 

this will ensure that the user has enough gas to increase the gas price a bit in the case of a gas price spike too

iamdavid0091:
Ha loving the usage of emojis in the quest details.

For the copy, it would be good to have it end with an action for the user. How about:

“Insufficient wallet balance to claim your reward due to gas fees. Try a different wallet or replenish this one before retrying.“

BrettCleary:
Alert icon is looking small 
image

src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/rewards/mintReward.ts
```diff
@@ -36,68 +36,89 @@ export async function mintReward({
...
   const depositContracts: DepositContract[] =
     await window.api.getDepositContracts(questId)
+
   const depositContractAddress = depositContracts.find(
     (val) => val.chain_id === reward.chain_id
   )?.contract_address
+
   if (depositContractAddress === undefined) {
-    console.error(
+    throw Error(
```
BrettCleary, line 45: ah you need this to trigger the mutation error. good to know

BrettCleary (commented):
working well for me. will review again after the copy changes are made 👍

src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
eliobricenov: good call

src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
eliobricenov: that is the average gas cost to execute the function, we can add estimated to the name

src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
flavioislima: Would be good to check with metamask, no? Maybe this is translated for some languages, I imagine for instance languages like Chinese, Korean or Russian or any other that don't use Latin chars @BrettCleary @eliobricenov

src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
eliobricenov: image

do you say we use the max?

BrettCleary (approved):
lgtm

jiyuu-jin (approved):
LGTM

nyghtstalker (approved):
Confirmed that I am seeing the errors messages appear when there is a problem claiming the rewards and there is not enough currency for gas.

image

image

PR #963, opened 2024-07-30
feat(quest): add support for sending completed claim to API

eliobricenov:
## Summary
After the claim transaction finishes, we need to send the transaction to the API to notify the reward was claimed and the signature was used.

Right now, as long as there is a valid signature, we show the playstreak reward as still not claimed.

## How to test
1. Claim an on-chain reward
2. Double check that the playstreak is reset in the UI after the balance reloads

closes https://github.com/HyperPlay-Gaming/product-management/issues/528

---


src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
```diff
@@ -165,6 +165,24 @@ export function QuestDetailsWrapper({
     }
   })
 
+  const confirmClaimMutation = useMutation({
+    mutationFn: async (params: ConfirmClaimParams) => {
+      return window.api.confirmRewardClaim(params)
+    },
+    retry: 5,
+    retryDelay: 1000,
```
eliobricenov, line 173: since this is a pretty critical flow I think is ok to add retries here

PR #971, opened 2024-07-30
[TECH] Check for HyperPlay games updates every hour

flavioislima:
This should fix the edge case of users having the app opened for a long period of time without hitting the refresh button as well.

---

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -676,6 +676,16 @@ if (!gotTheLock) {
 
     initTrayIcon(mainWindow)
 
+    // Call checkGameUpdates for HyperPlay games every hour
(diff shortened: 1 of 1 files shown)


src/backend/main.ts
```diff
@@ -676,6 +676,16 @@ if (!gotTheLock) {
 
     initTrayIcon(mainWindow)
 
+    // Call checkGameUpdates for HyperPlay games every hour
+    const checkGameUpdatesInterval = 1 * 60 * 60 * 1000
```
flavioislima, line 680: What do you guys think about the interval? should be higher or it is fine?

flavioislima: well, lets keep 1h for now and then change in the future if we need.

PR #976, opened 2024-07-30
[Fix] long reward title

BrettCleary:
# summary
- fix long reward titles overflowing


eliobricenov:
@BrettCleary I think some changes are missing? The only change is in the yarn lock file

PR #984, opened 2024-07-31
[Feat] Add window.api.navigate to store preload, add quest routes

BrettCleary:
# Summary
- add generic api navigate function to store preload with `NavigateListener` frontend component handler so that it can be called at any time
- add `quests/:questId` route
- replace `setSelectedQuestId` with react-router-dom `navigate` and the path param `:questId`


jiyuu-jin (approved):
LGTM

eliobricenov (approved):

nyghtstalker (approved):
Clicking any of the quests on the store page now take the user to the quests page instead of the store quests page.

PR #985, opened 2024-07-31
feat(quest): add claim flags for each reward type

eliobricenov:
Add a flag for each reward type to control when they can be claimed


src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
```diff
@@ -380,10 +389,17 @@ export function QuestDetailsWrapper({
...
+        (reward) => rewardTypeClaimEnabled[reward.reward_type]
+      )
+    )
+
     const ctaDisabled =
       !flags.questsOverlayClaimCtaEnabled ||
       (!isEligible() && !showResyncButton) ||
-      isClaiming
+      isClaiming ||
+      !isRewardTypeClaimable
```
BrettCleary, line 402: we'll have to keep this in mind when we add multiple rewards per quest. that the cta button for claim all will be enabled even if we only want claim to be enabled for 1 of the n reward types for that quest

might be worth adding a TODO here for that

BrettCleary (approved):
tested all the flag toggles except for 1155, all working for me. great job!

jiyuu-jin (approved):

PR #995, opened 2024-08-02
[FIX] Deal with Epic games listing from Quests Page

### LINKED ISSUE #990: Epic forwarding games are added directly to library when navigating through the Quests tab
### Describe the bug

User are able to add Epic forwarding games to their library via the Quests tab. 

### Add logs

N/A

### Steps to reproduce

1. Go to the store game page for Apeiron
2. Observe that it is Epic forwarding (Download on Epic)
3. Go to the Quests tab
4. Select "Game7 Credits" Apeiron" quest
5. Click View Game
6. Observe that the user is taken to the "HyperPlay" version of Apeiron and not the Epic game

### Expected behavior

Need to figure out the best way to handle "View Game" for quests that have Epic forwarding games

### Screenshots

image
image
image
image

### hyperplay Version

Latest Stable

### System Information

- Windows 10
- hyperplay Version 0.17.2

### Additional information

_No response_

flavioislima:
FIX #990 by  adding some methods to get the epic url from the API and checking it with the epic games on the use library then navigating either to the library page (when the game is on the library) or to the epic games store page.

---

### DIFF
--- a/src/backend/storeManagers/hyperplay/library.ts
+++ b/src/backend/storeManagers/hyperplay/library.ts
@@ -18,7 +18,6 @@ import {
 } from './utils'
 import { getGameInfo as getGamesGameInfo } from './games'
 import { getValistListingApiUrl, qaToken } from 'backend/constants'
-import { ProjectMetaInterface } from '@valist/sdk/dist/typesShared'
 
 export async function addGameToLibrary(projectId: string) {
   const currentLibrary = hpLibraryStore.get('games', [])
@@ -45,38 +44,6 @@ export async function addGameToLibrary(projectId: string) {
   const data = res.data
   const gameInfo = getGameInfoFromHpRelease(data)
   hpLibraryStore.set('games', [...currentLibrary, gameInfo])
-}
-
-// hit the valist api and get the info if the game is a epic listing
-export async function getEpicListingUrl(projectId: string): Promise<string> {
-  const listingUrl = getValistListingApiUrl(projectId)
-
-  try {
-    const getConfig =
-      qaToken !== '' ? { headers: { Authorization: `Bearer ${qaToken}` } } : {}
-    const res = await axios.get<HyperPlayRelease>(listingUrl, getConfig)
-
-    if (res.status !== 200) {
-      logError(`Error when trying to get listing info ${res}`)
-      return ''
-    }
-
-    const data = res.data
-    const { epic_game_url, launch_epic } =
-      data.project_meta as ProjectMetaInterface
-
-    if (
-      !!launch_epic &&
-      !!epic_game_url &&
-      epic_game_url.startsWith('https://store.epicgames.com/')
-    ) {
-      return epic_game_url
-    }
-    return ''
-  } catch (error) {
-    logError(`Error when trying to get listing info: ${error}`)
-    return ''
-  }
 }
 
 export const getInstallInfo = async (
--- a/src/backend/storeManagers/hyperplay/library.ts
+++ b/src/backend/storeManagers/hyperplay/library.ts
@@ -18,6 +18,7 @@ import {
 } from './utils'
 import { getGameInfo as getGamesGameInfo } from './games'
 import { getValistListingApiUrl, qaToken } from 'backend/constants'
+import { ProjectMetaInterface } from '@valist/sdk/dist/typesShared'
 
 export async function addGameToLibrary(projectId: string) {
   const currentLibrary = hpLibraryStore.get('games', [])
@@ -44,6 +45,38 @@ export async function addGameToLibrary(projectId: string) {
   const data = res.data
   const gameInfo = getGameInfoFromHpRelease(data)
   hpLibraryStore.set('games', [...currentLibrary, gameInfo])
+}
+
+// hit the valist api and get the info if the game is a epic listing
+export async function getEpicListingUrl(projectId: string): Promise<string> {
+  const listingUrl = getValistListingApiUrl(projectId)
+
+  try {
+    const getConfig =
+      qaToken !== '' ? { headers: { Authorization: `Bearer ${qaToken}` } } : {}
+    const res = await axios.get<HyperPlayRelease>(listingUrl, getConfig)
+
+    if (res.status !== 200) {
+      logError(`Error when trying to get listing info ${res}`)
+      return ''
+    }
+
+    const data = res.data
+    const { epic_game_url, launch_epic } =
+      data.project_meta as ProjectMetaInterface
+
+    if (
+      !!launch_epic &&
+      !!epic_game_url &&
+      epic_game_url.startsWith('https://store.epicgames.com/')
+    ) {
+      return epic_game_url
+    }
+    return ''
+  } catch (error) {
+    logError(`Error when trying to get listing info: ${error}`)
+    return ''
+  }
 }
 
 export const getInstallInfo = async (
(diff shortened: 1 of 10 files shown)


src/backend/storeManagers/hyperplay/library.ts
```diff
@@ -46,6 +47,38 @@ export async function addGameToLibrary(projectId: string) {
   hpLibraryStore.set('games', [...currentLibrary, gameInfo])
 }
 
+// hit the valist api and get the info if the game is a epic listing
+export async function getEpicListingUrl(projectId: string): Promise<string> {
```
BrettCleary, line 51: I think this should be in `src\backend\storeManagers\hyperplay\utils.ts` since we try to keep the store manager library and games modules' api's 1:1 with all the other stores

src/frontend/helpers/index.ts
```diff
@@ -148,6 +149,41 @@ export function getPlatformName(platform: string): string {
   }
 }
 
+const getLastPartOfUrl = (url: string) => {
+  return url.split('/').pop()
+}
+
+export const fetchEpicListing = async (projectId: string) => {
```
BrettCleary, line 156: we'll just need to remember to qa this for new epic game forwarder listings since developers could enter an epic url that might break this

src/frontend/screens/Quests/components/QuestDetailsViewPlay/index.tsx
```diff
@@ -36,11 +37,38 @@ export function QuestDetailsViewPlayWrapper({
...
-      navigate(`/gamepage/hyperplay/${appName}`, {
-        state: { gameInfo, fromDM: false }
-      })
+      const { appName: epicAppName, epicListingUrl } = await fetchEpicListing(
+        appName
+      )
+      const runner: Runner = epicListingUrl ? 'legendary' : 'hyperplay'
+
+      // check for gameinfo to see if it is on the library
+      return getGameInfo(epicAppName || appName, runner)
```
BrettCleary, line 46: can we set these in the same if statement like
```
let runner = 'hyperplay'
let name = appName
if (epicListingUrl && epicAppName){
  runner = 'legendary
  name = epicAppName
}
```

otherwise if `fetchEpicListing` returns empty string for appName but a correct epic listing url, it would try to navigate to the 'hyperplay' runner with the epic app name

src/frontend/screens/Quests/components/QuestDetailsViewPlay/index.tsx
```diff
@@ -36,11 +37,38 @@ export function QuestDetailsViewPlayWrapper({
...
+        .then((res) => {
+          if (!res) {
+            throw new Error('Game not found in library')
+          }
+
+          return navigate(
+            `/gamepage/${runner}/${epicAppName ? epicAppName : appName}`,
```
BrettCleary, line 53: can reference the same variable for app name here too

BrettCleary (commented):
I am unable to launch this branch in dev mode

I get this error `Cannot use 'in' operator to search for 'CallExpression' in undefined`

src/frontend/screens/Quests/components/QuestDetailsViewPlay/index.tsx
BrettCleary, line 1: I think  the logic you added for navigation here could be abstracted away to a hook too so that we can reuse this functionality elsewhere on the frontend

BrettCleary (approved):
It works but I get 100's of lines logged to the console on click. it logs an axios response after

```
(15:06:55) ERROR:   [Legendary]:        Error Getting Info from Epic API
(15:06:55) ERROR:   [Legendary]:        AxiosError: Request failed with status code 403
... 100's of lines of axios response logs
```

we should clean up this logging

nyghtstalker:
**-- When the user IS logged into EPIC:**

I believe it was the initial attempt to install the game from the quests page I got this warning on the install modal

image

However, after every other attempt, I could not replicate this and this warning doesn't appear again.

Aperion is the only game in quests, atm, that has an EPIC forward. The user is now taken to the EPIC game page instead of the HP game page. 


**-- When the user IS NOT logged into EPIC:**

Clicking the game via the quests page now takes the user to the Epic Store game page.

image

With this flow the user needs to have (or will need to create) an EPIC account and then log into the account via HP in order to play it. Once the user is logged in the user will need to know how to access/install the game since this store page does not trigger a download or take the user to the install page. 

At this point once the user adds the game to their EPIC library they have to wait for it to appear in their EPIC library and then find and install it via the EPIC page in the Library. To my knowledge this is the only way to access an EPIC game. It is fairly complicated and if the user does not know what to do it is very difficult.

With all that being said...I feel the best option is to either now allow EPIC forward games to have quests (since this is process is complicated and I think it will cause user frustration) OR they need to have a version uploaded to HP (which I think nearly every game does submit one via the review process) and they use that version instead of forwarding to EPIC.

I know this is a lot but I think this needs to get addressed and figured out to help improve the user experience. @BrettCleary @flavioislima @awantoch

BrettCleary:
@flavioislima is there a way we can programmatically add an epic game to the user's library with legendary? we can handle on a separate branch/issue but I think this would address @nyghtstalker's concerns. curious if it's possible.

To your point @nyghtstalker, Heroic has been using this pattern for a long time, and while steps can be consolidated, it should work for now. We will always be able to provide better support for games directly on HP store and game devs understand that.

flavioislima:
@nyghtstalker 

1. This one was a false positive, might happens from time to time, not related with the PR. If the user clicks install and they had permission it will work just fine. i can double check this method on another PR later.
2. Unfortunately there is not much we can do about it to improve this workflow. The Epic integration is implemented as it is and we only read the list of games the user already has in their library. to make it worse it might take a few minutes for the game to appear on HyperPlay library and sometimes even a restart. In this matter we are pretty limited.
3. About quests for Epic games this should be something we need to discuss for sure, or find ways of improving the flow by adding this type of information/instructions to the quest itself. Let's bring that to our next team meeting.
4. @BrettCleary Legendary only reads information from Epic, there is not way to add a game to the user library without going through the Epic website. Even for free games we need to go through their carting system. In Heroic there were some researches on how to add the weekly free game to the user library but without much success.

src/frontend/helpers/index.ts
flavioislima: Yes. The only way to make this 100% reliable would be to have the Epic `appName` in the dev portal. For every new listing with quests we will need to double check this and even adjust it.

src/frontend/screens/Quests/components/QuestDetailsViewPlay/index.tsx
flavioislima: Good call

src/backend/storeManagers/hyperplay/library.ts
flavioislima: done

src/frontend/screens/Quests/components/QuestDetailsViewPlay/index.tsx
flavioislima: Ok, after changing it stoppeed working and now I remember why I did this way. It is because for the last case (navigate to epic store) we need it to be runner legendary BUT has an epic listing as well.
but I can adjust this logic.

jiyuu-jin (approved):

nyghtstalker (approved):
@flavioislima @BrettCleary Thank you both for all the feedback. I def agree that we need to figure out the best flow for a better experience with all this.  I will make a card to help track for an improvement on this. This PR is working as intended and I won't let this hold it up. Great job on this, sir!

PR #997, opened 2024-08-02
[Feat] add search bar to quest summary table

BrettCleary:
# Summary
- add search bar to quest summary table for searching quests on quest title or game title

# Screenshots
image
image

### DIFF
--- a/src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
+++ b/src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
@@ -387,6 +387,7 @@ export function QuestDetailsWrapper({
   if (
     selectedQuestId !== null &&
     questMeta !== undefined &&
+    questMeta !== null &&
     questRewards !== undefined
   ) {
     const isRewardTypeClaimable = Boolean(
--- a/src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
+++ b/src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
@@ -386,9 +386,8 @@ export function QuestDetailsWrapper({
 
   if (
     selectedQuestId !== null &&
-    questMeta !== undefined &&
-    questMeta !== null &&
-    questRewards !== undefined
+    questMeta &&
+    questRewards
   ) {
     const isRewardTypeClaimable = Boolean(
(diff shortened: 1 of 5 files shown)


red-game-dev (approved):
LGTM to me, checked the code as still need to clone client repo and set it up (I think i need .env) 🙌 Awesome work! Approved on the code check

src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
```diff
@@ -387,6 +387,7 @@ export function QuestDetailsWrapper({
   if (
     selectedQuestId !== null &&
     questMeta !== undefined &&
+    questMeta !== null &&
     questRewards !== undefined
```
eliobricenov, line 391: I think this can be `selectedQuestId !== null && questMeta && questRewards`.

src/frontend/hooks/useGetQuest.ts
```diff
@@ -4,7 +4,7 @@ import { useQuery, useQueryClient } from '@tanstack/react-query'
 export default function useGetQuest(questId: number | null) {
   const queryClient = useQueryClient()
   const queryKey = `getQuest:${questId}`
-  const query = useQuery<Quest>({
+  const query = useQuery<Quest | null>({
```
eliobricenov, line 7: I don't think we need this, react-query already infers the value, no need to use the type

eliobricenov (approved):
lgtm left some nitpicks

src/frontend/hooks/useGetQuest.ts
BrettCleary: without this, I wasn't getting the linting/autocomplete previously. looks like it is working now on my machine without specifying the type though. will make the change 👌

nyghtstalker:
1. Search for and then select from the results. Clear the results and then search again. Observe that the results do not auto-appear.

image
image
image


2. With a quest opened search for something and then shrink the panel. The search results will be on top and still display.

image
image

nyghtstalker (approved):
Everything works great!

My feedback will be addressed at a later date. 

Quest search results do not auto-appear after clearing https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/issues/1005
Quest search results still appear and is on top of the page https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/issues/1006

PR #998, opened 2024-08-03
[UX] sync playsession every minute

BrettCleary:
# Summary
This PR syncs the play session every minute while playing a game.

closes https://github.com/HyperPlay-Gaming/product-management/issues/519

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -1155,6 +1155,9 @@ async function syncPlaySession(appName: string, runner: Runner) {
     (stopPlayingTimeMonotonic - gamePlaySessionStartTimes[appName]) /
     BigInt(1000000)
 
+  // reset the time counter
+  startNewPlaySession(appName)
+
   // update local json with time played
   const sessionPlaytimeInMinutes =
     sessionPlaytimeInMs / BigInt(1000) / BigInt(60)
@@ -1178,9 +1181,6 @@ async function syncPlaySession(appName: string, runner: Runner) {
       parseInt((sessionPlaytimeInMs / BigInt(1000)).toString())
     )
   }
-
-  // reset the time counter
-  startNewPlaySession(appName)
 
   return sessionPlaytimeInMs
 }
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -1141,6 +1141,56 @@ ipcMain.on('logError', (e, err) => logError(err, LogPrefix.Frontend))
 ipcMain.on('logInfo', (e, info) => logInfo(info, LogPrefix.Frontend))
 
 let powerDisplayId: number | null
+const gamePlaySessionStartTimes: Record<string, bigint> = {}
+
+function startNewPlaySession(appName: string) {
+  // Uses hrtime for monotonic timer not subject to clock drift or sync errors
+  const startPlayingTimeMonotonic = hrtime.bigint()
+  gamePlaySessionStartTimes[appName] = startPlayingTimeMonotonic
+}
+
+async function syncPlaySession(appName: string, runner: Runner) {
+  const stopPlayingTimeMonotonic = hrtime.bigint()
+  const sessionPlaytimeInMs =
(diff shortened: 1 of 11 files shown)


red-game-dev (approved):
LGTM 🚀

src/backend/main.ts
```diff
@@ -1141,6 +1141,56 @@ ipcMain.on('logError', (e, err) => logError(err, LogPrefix.Frontend))
...
+
+  const totalPlaytime =
+    sessionPlaytimeInMinutes + BigInt(tsStore.get(`${appName}.totalPlayed`, 0))
+  tsStore.set(`${appName}.totalPlayed`, Number(totalPlaytime))
+
+  const game = gameManagerMap[runner].getGameInfo(appName)
+  const { gameIsEpicForwarderOnHP, hyperPlayListing } =
+    await gameIsEpicForwarderOnHyperPlay(game)
+
+  if (gameIsEpicForwarderOnHP && hyperPlayListing) {
```
jiyuu-jin, line 1173: Are we sending this for all Epic Forwarder games and all HyperPlay games even if they don't have a play session?

src/backend/main.ts
```diff
@@ -1141,6 +1141,56 @@ ipcMain.on('logError', (e, err) => logError(err, LogPrefix.Frontend))
...
+    postPlaySessionTime(
+      hyperPlayListing.project_id,
+      parseInt((sessionPlaytimeInMs / BigInt(1000)).toString())
```
jiyuu-jin, line 1176: Couldn't you do something like `hyperPlayListing?.project_id || appName` and combine these lines.

src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
```diff
@@ -178,6 +182,20 @@ export function QuestDetailsWrapper({
       loading: val.isLoading || val.isFetching
     })) ?? []
 
+  useEffect(() => {
```
jiyuu-jin, line 185: Might be able to break these useEffects into hooks to make this a bit more readable/maintainable, maybe something like:

```ts
import { useEffect } from 'react';

// Custom hook for tracking events
function useTrackEvent(selectedQuestId: number | null) {
  useEffect(() => {
    if (selectedQuestId !== null) {
      window.api.trackEvent({
        event: 'Quest Viewed',
        properties: { quest: { id: selectedQuestId.toString() } },
      });
    }
  }, [selectedQuestId]);
}

// Custom hook for sync interval
function useSyncInterval(projectId: string, selectedQuestId: number | null, invalidateQuery: () => void) {
  useEffect(() => {
    const syncTimer = setInterval(async () => {
      await window.api.syncPlaySession(projectId, 'hyperplay');
      // allow for some time before read
      await wait(5000);
      await invalidateQuery();
      resetSessionStartedTime();
    }, 1000 * 60);

    return () => {
      clearInterval(syncTimer);
    };
  }, [projectId, selectedQuestId, invalidateQuery]);
}
```

jiyuu-jin (approved):

eliobricenov (approved):
code lgtm

src/backend/main.ts
BrettCleary: well. they will always have a play session by this point. I think you're asking if the game doesn't have an active play streak quest and yes that's correct. We send this every time a game is played. 

When we add the quest id's to the listings endpoint, we can add this info to the `GameInfo` object and then check before calling this from the 'launch' handler.

src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
BrettCleary: made the refactor 👍

src/backend/main.ts
BrettCleary: yup. this simplifies the code a bit

nyghtstalker (approved):
It took a second after 100% for the 1/7 and bolt to appear but it worked!

image
image
image

PR #1000, opened 2024-08-06
[TECH] Auto-Update improvements

flavioislima:
### Changes

- Check for updates every hours instead of 12h.
- Download in the background and show only a single Dialog asking to restart.
- If Yes, apply update then restart.
- If No, apply the update after closing the app and reopening.
- Fix showing two dialogs on every update.

## How to Test

To test auto update on Windows locally make sure you added the option `"verifyUpdateCodeSignature": false`
under `build.win` in package.json  as bellow and change the app version to an old number, eg.: 0.17.1 and build it with `yarn dist:win` from the terminal.

Install the build version from the dist folder and check if it works.

package.json snippet:
```
    "win": {
      "artifactName": "${productName}-${version}-Setup-${arch}.${ext}",
      "icon": "build/win_icon.ico",
      "asarUnpack": [
        "build/bin/win32/legendary.exe",
        "build/bin/win32/gogdl.exe"
      ],
      "files": [
        "build/bin/win32/*"
      ],
      "verifyUpdateCodeSignature": false
    },
```

AND

` "version": "0.17.1"`

---


BrettCleary:
> If No, apply the update after closing the app and reopening.

This part is not working. I 
1. built this as 0.17.1. 
2. launched
3. got this popup
image
4. clicked no
5. closed the app
6. reopened
7. the app was still 0.17.1

src/backend/updater.ts
```diff
@@ -7,10 +7,10 @@ import { logError, LogPrefix, logInfo } from './logger/logger'
 import { isOnline } from './online_monitor'
 
 autoUpdater.autoDownload = true
-autoUpdater.autoInstallOnAppQuit = true
+autoUpdater.autoInstallOnAppQuit = false
```
BrettCleary, line 10: don't we need this true?

src/backend/updater.ts
```diff
@@ -7,10 +7,10 @@ import { logError, LogPrefix, logInfo } from './logger/logger'
...
 
-// check for updates every 12 hours
-const interval = 1000 * 60 * 60 * 12
+// check for updates every 6 hours
+const interval = 1000 * 60 * 60 * 6
```
BrettCleary, line 13: is there a downside to making this shorter? might be useful if we need to get out an update fast. maybe 1 hr?

BrettCleary:
If I click Yes, HyperPlay does not restart and the app is not updated
image

BrettCleary (commented):
not working for me on windows but instead of two dialog boxes showing, it's only 1 so that's progress

flavioislima:
> not working for me on windows but instead of two dialog boxes showing, it's only 1 so that's progress 

Hmm, I'm not 100% sure that is testable though.
Need to double check in main because since the certificates will be different it should fail 🤔

src/backend/updater.ts
flavioislima: No reason, I can make it shorter

src/backend/updater.ts
flavioislima: Yeah, it changes if the user clicks NO just bellow.

flavioislima:
I will update the `electron-updater` package to see if it will improve something, but I cannot update the app when building from `main` as well. it gives me ENOENT because it fails to download the app.

flavioislima:
Added a comment on how to test locally on windows.

eliobricenov (approved):
code lgtm

jiyuu-jin (approved):
LGTM, loving all the extra logs

BrettCleary (changes requested):
Auto update on windows is giving the following errors after re-launching the app
image
image

flavioislima:
> Auto update on windows is giving the following errors after re-launching the app image image

Can you give more context? How do I reproduce it?

nyghtstalker:
- Confirmed that "Check for HyperPlay Updates on Startup" setting must be checked for the white update modal to appear. If it is not checked it will not appear.
- Confirmed that only one white update modal appears. 
- Clicking No closes the modal. 
- Clicking Yes closes the client and starts the installer. Once the installer is complete it does update to 0.18.1.

Anything else you can think of that needs to be checked @flavioislima ?

image
image

image

nyghtstalker:
Seeing these on Mac. Clicking Download triggers the second image to appear. Clicking YES takes the user to the HP Downloads page.

image

flavioislima:
> Results from testing on Mac.
> 
> * Only this modal appears after launch (the client does not open). This opens BEHIND windows so very easy to miss. This should appear on TOP of everything so it is known and doesn't need to be searched for.
>   image
> * Clicking either Download or Cancel triggers this modal (client still not launched)
>   image
> * Clicking YES takes the user to the HP Downloads page and then opens the client. No auto-download.
> * Clicking NO opens the client.

Ah yeah, I forgot to say that I only changed the build to be tested on Windows. I will see if there is a way to test on mac as well, some flag on the docs.

nyghtstalker (approved):
Mac issues will be addressed in a different PR. I created this ticket for reference.
https://github.com/HyperPlay-Gaming/product-management/issues/585

Since this is good on Windows this is G2G.

flavioislima:
> Mac issues will be addressed in a different PR. I created this ticket for reference.
> https://github.com/HyperPlay-Gaming/product-management/issues/585
> 
> Since this is good on Windows this is G2G.

Mac will also be fixed by this PR just to make it clear.
But we will only know after two releases since this code will only take effect on the next one and then will only be triggered when we do another release after the next one.

Let's just keep this in mind since after the next release people will still have issues.

BrettCleary (approved):
working on windows 👌

PR #1002, opened 2024-08-06
[Fix] should show extension

BrettCleary:
# Summary
Fixes extension showing if connection method changes after overlay is launched

closes https://github.com/HyperPlay-Gaming/product-management/issues/524

# Testing
https://app.qase.io/project/HC?case=355&suite=5
https://app.qase.io/project/HC?case=354&suite=5

### DIFF
--- a/src/frontend/OverlayManager/Overlay/index.tsx
+++ b/src/frontend/OverlayManager/Overlay/index.tsx
@@ -56,7 +56,7 @@ export const Overlay = observer(function ({
   }
 
   const shouldShowExtension =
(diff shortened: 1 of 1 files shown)


eliobricenov (approved):

red-game-dev (approved):

nyghtstalker (approved):
Confirmed that the user is now able to unlock wallet in a browser game if launched with the wallet locked.

Confirmed that the user can use the extension after switching when launching the game with a different wallet.

PR #1003, opened 2024-08-06
feat(quest): add onchain reward confirmation modal 

eliobricenov:
closes https://github.com/HyperPlay-Gaming/product-management/issues/532

how to test:

1. try to claim an on-chain request
2. check that the confirmation modal is shown
3. check that the claim flow is resumed after hitting confirm

---


src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
```diff
@@ -89,6 +93,63 @@ async function getRewardClaimGasEstimation(reward: Reward) {
...
 }
 
+interface ConfirmProps extends ModalProps {
+  networkName: string
+  onConfirm: () => void
+  onCancel: () => void
+}
+
+function ConfirmClaimModal(props: ConfirmProps) {
+  const { t } = useTranslation()
```
BrettCleary, line 103: should we move this to a separate components folder?

eliobricenov: I don't think so, this component will only be used here so I think it is better to keep them close.

src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
```diff
@@ -89,6 +93,63 @@ async function getRewardClaimGasEstimation(reward: Reward) {
...
+            <span className="text--bold">
+              {t('quest.claimWarning.important', 'IMPORTANT')}:
+            </span>{' '}
+            {t(
+              'quest.claimWarning.networkMessage',
+              `Please ensure that you are allocating enough gas on the {{networkName}} network for the transaction to be successfully confirmed`,
+              { networkName: props.networkName }
+            )}{' '}
+            <span className="text--bold">
+              {t('quest.claimWarning.within24Hrs', 'within 24 hrs')}
```
BrettCleary, line 126: this will have issues translating. 'react-i18next' has a `Trans` component for this so that you can keep the translation string together https://stackoverflow.com/questions/75309102/how-to-add-a-partial-style-for-a-i18n-translation

since in other languages each component of the sentence is not always in the same order

src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
```diff
@@ -89,6 +93,63 @@ async function getRewardClaimGasEstimation(reward: Reward) {
...
+          </span>
+          <span>
+            {t('quest.claimWarning.otherwise', 'Otherwise, the Quest Reward')}{' '}
+            <span className="text--bold">
+              {t(
+                'quest.claimWarning.expire',
+                'will expire and will no longer be claimable'
+              )}
+              .
+            </span>
```
BrettCleary, line 137: we could do the same here

BrettCleary (changes requested):
see my comments above. just a couple minor change requests. haven't tested yet

public/locales/en/translation.json
```diff
@@ -1,959 +1,966 @@
-{
```
eliobricenov, line 1: all these changes are because I ran `yarn i18n`

flavioislima: You probably edited the file manually and then saved and it changed the formatting. We should never edit the json file from translations directly, always use the i18n command when adding or removing new keys.

BrettCleary:
The letter C in the CTA's should be capitalized 
image

src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
BrettCleary: this still goes against the pattern we use elsewhere. If this component is the only one that uses it, we would normally just put it in a subfolder alongside this component

BrettCleary (approved):
code LGTM. modal works

we should address the CTA capitalization before merging and if you have time, it'd be good to move the modal to a component in a subfolder

flavioislima (approved):
LGTM

src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
eliobricenov: true!

nyghtstalker (approved):
Confirmed that this modal appears after clicking "Claim all" on ERC20 and ERC721 quests. If you are not on Sepo it does request the change and switch you. If you attempt to claim again it does give an error which is expected.

image

PR #1013, opened 2024-08-08
[TECH] Upgrade Electron and other Dev Dependencies

flavioislima:
### A full-regression test would be good for this one although there were no major breaking changes on any of those packages.

Updated mostly to mid/minor versions. Just a few ones upgraded to major.
Node types updated to match the version in electron 32 (node 20).

### News 
- Electron: https://releases.electronjs.org/release/v32.0.0
- i18n-parser: Now it will update current keys if we change the default value in the code, easier then removing and re-adding it like before. Not enabled though since it replaced current translations with empty value due to those having no default value. We can fix that on another PR and then enable the setting. Changed EOL to `match` instead of `auto` to see if it will fix current issue with it changing depending on the OS on our PRs.
- Husky: simplified scripts writing

### Could not upgrade
- Typescript: Will require their own PRs due to the amount of changes required. Also, maybe doesn't worth it, should be discussed on a dev meeting.
- vite-plugin-svgr: simply breaks on any version over 2.2.2. Tried adding `?react` at the end of the imports like they say in the docs, tried changing TS settings and adding options in `vite.config` file without success. Didn't want to put more effort here since it is working fine with current vite version and it is not a major dependency or with security issues.

---


eliobricenov (approved):

red-game-dev (approved):

jiyuu-jin (approved):

src/backend/main.ts
```diff
@@ -1127,13 +1127,13 @@ if (existsSync(installed)) {
   })
 }
 
-ipcMain.handle('refreshLibrary', async (e, library?) => {
+ipcMain.handle('refreshLibrary', async (e, library?: Runner | 'all') => {
```
BrettCleary, line 1130: this shouldn't be needed. it is inferred from the ipcBridge.d.ts declaration

src/frontend/state/AchievementState.ts
```diff
@@ -77,7 +77,6 @@ class AchievementState {
       {},
       {
         deep: true,
-        proxy: false,
```
BrettCleary, line 80: why was this removed?

src/frontend/SentryHandler/index.tsx
```diff
@@ -19,6 +19,7 @@ const SentryHandler = () => {
             ? devSentryDsn
             : prodSentryDsn
       },
+      // @ts-expect-error - not sure why is this failling
       reactInit
```
BrettCleary, line 23: might be a mismatch between the `@sentry/electron` and `@sentry/react` package versions?

BrettCleary (approved):
agreed. we likely need a full regression test on this one

src/frontend/state/AchievementState.ts
flavioislima: It says that the property doesn't exists so gives eslint error.
Maybe was removed on the latest mobx version.

src/frontend/SentryHandler/index.tsx
flavioislima: I will check

src/backend/main.ts
flavioislima: ok, this was showing an error before and now its not so maybe was some typescript cache, not sure.

src/frontend/SentryHandler/index.tsx
flavioislima: I think maybe we should update Sentry to latest packages and do the SDK migration on another PR.
I was looking into the migration guide and the new docs and they diverge in some points.
The electron one says its possible to use the Sentry.init to start a second SDK, but on the latest version its not possible anymore because it just accepts the options so on the frontend we would use only the React SDK instead of also importing from sentry/electron/renderer.
Since I don't really know how to test this lets keep it like this on this PR.

But that complain that I added the expect error its because it says that the type of ReactInit mismatches what Sentry.init is expecting.

src/frontend/state/AchievementState.ts
flavioislima: so I added this back and then removed it again after reading the docs:

> proxy: false forces observable(thing) to use non-[proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) implementation. This is a good option if the shape of the object will not change over time, as non-proxied objects are easier to debug and faster. **This option is not available for make(Auto)Observable**, see [avoiding proxies](https://mobx.js.org/observable-state.html#avoid-proxies).

So Typescript was not wrong in the end and this method has no `proxy` option, only the `observable` does.

nyghtstalker (approved):
Looks good after full regression. No issues found. Great job!

PR #1032, opened 2024-08-19
[Tech] update playsession

BrettCleary:
# Summary

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -1144,15 +1144,20 @@ ipcMain.on('logError', (e, err) => logError(err, LogPrefix.Frontend))
 ipcMain.on('logInfo', (e, info) => logInfo(info, LogPrefix.Frontend))
 
 let powerDisplayId: number | null
-const gamePlaySessionStartTimes: Record<string, bigint> = {}
+let gamePlaySessionStartTimes: Record<string, bigint> = {}
 
 function startNewPlaySession(appName: string) {
+  gamePlaySessionStartTimes = {}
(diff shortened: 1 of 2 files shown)


src/backend/main.ts
```diff
@@ -1144,15 +1144,20 @@ ipcMain.on('logError', (e, err) => logError(err, LogPrefix.Frontend))
 ipcMain.on('logInfo', (e, info) => logInfo(info, LogPrefix.Frontend))
 
 let powerDisplayId: number | null
-const gamePlaySessionStartTimes: Record<string, bigint> = {}
+let gamePlaySessionStartTimes: Record<string, bigint> = {}
 
 function startNewPlaySession(appName: string) {
+  gamePlaySessionStartTimes = {}
```
jiyuu-jin, line 1150: I see, you just clear the object when another game is launched. Why not just make it a single gamePlaySessionStartTime?

jiyuu-jin (approved):

red-game-dev (approved):
Since it is urgent matter, did review the code 🙌 LGTM 🚀

nyghtstalker:
1. I am able to launch Bushi and then Craft World after at the same time.

image

2. While Market Wars is launched Bushi does not open the game but the button does display "Playing" which def seems like it should be running and it is broken.

At the very least I feel that button shouldn't continue to say Playing and/or we should add a notification stating that two games can't be launched at once.

image

nyghtstalker (approved):
This looks good! I tried to break it (and the UI/UX does not work) but it seems like the quest progress is stopped for the first game launched!

PR #1036, opened 2024-08-21
fix(quest-claim): balance check race condition

eliobricenov:
fix the race condition of the wallet balance check after switching account.

## Summary

This PR fixes a bug related to the `useBalance` hook when switching chains. The issue was that after `switchChainAsync` resolved, the balance was not updated to reflect the current chain's balance. This caused an incorrect balance check when a user switched to the quest's network, potentially allowing a signature to be granted even if the user didn't have enough gas on the new chain.

To resolve this, the balance is now fetched directly using the [`getBalance`](https://wagmi.sh/core/api/actions/getBalance#chainid) function from the WAGMI API, ensuring that the balance reflects the correct chain immediately after switching, without waiting for WAGMI to fully update the chain configuration in the hooks.

## How to Test

1. Find a quest with a valid playstreak for your account.
2. Ensure your wallet **is not** on the quest's network but has enough gas for a claim.
3. Click "Claim."
4. Verify that after the network switches, the balance check is accurate and prevents the claim if there isn't enough gas on the new network.

You should see a log like this after step 4:
```
(17:00:09) ERROR:   [Frontend]:         Not enough balance in the connected wallet to cover the gas fee associated with this Quest Reward claim. Current balance: 32773032999000, gas needed: 93089192252551
```

---

### DIFF
--- a/src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
+++ b/src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
@@ -30,6 +30,8 @@ import { InfoAlertProps } from '@hyperplay/ui/dist/components/AlertCard'
 import { useSyncPlaySession } from 'frontend/hooks/useSyncInterval'
 import { useTrackQuestViewed } from 'frontend/hooks/useTrackQuestViewed'
 import { ConfirmClaimModal } from './components/ConfirmClaimModal'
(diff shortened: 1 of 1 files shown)


jiyuu-jin (approved):

nyghtstalker (approved):
Confirmed that while on a different chain than what is needed to claim a reward with not enough gas that the network is automatically switched to the correct network and the insufficient balance error is displayed!

image
image

PR #1037, opened 2024-08-21
feat(quest-claim): show error ui for claim errors

eliobricenov:
We were try-catching the errors in the claim functions. We need to let the functions throw so they can be caught by the mutation. It will handle it and store the value in `mutation.error`.

```ts
if (writeContractError || claimRewardsMutation.error || switchChainError) {
 // show UI error
}
```

---


eliobricenov:
With the current code if there is an error, mutation.error will be undefined and we won't show the error UI.

BrettCleary (approved):
LGTM

jiyuu-jin (approved):

nyghtstalker (approved):
Confirmed that after attempting to claim a reward followed by a reject a different account is unable to claim the same reward.

image
image

PR #1040, opened 2024-08-22
[Fix] repeat external task claims showing sync instead of claim

BrettCleary:
# Summary
This PR
- invalidates user playstreak queries on claim points, claim tasks, resync task queries
- does not show the sync button if the streak has been met even if completed_counter > 0

# Testing

1. open a g7 credits quest with 7 day streak
2. see claim all
3. click it
4. it changes to sync and streak resets to 0
5. complete streak again
6. see that the button is now claim all again
7. click it
8. it changes to sync and streak resets to 0


src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
```diff
@@ -212,8 +221,16 @@ export function QuestDetailsWrapper({
   const userId = session.data?.userId
   const isSignedIn = !!userId
 
+  const hasMetStreak =
+    questPlayStreakData?.current_playstreak_in_days !== undefined &&
+    questMeta?.eligibility?.play_streak?.required_playstreak_in_days !==
+      undefined &&
+    questPlayStreakData?.current_playstreak_in_days >=
+      questMeta.eligibility?.play_streak?.required_playstreak_in_days
+
```
eliobricenov, line 230: I've always wondered how to handle optional numeric values because I know `0` is falsy, but I've always like to do boolean checks to catch both `undefined` and `null` without having to do 

```js
if (value !== undefined && value !== null)
```

You can use the chain operator for this, although I can agree it looks funny :D 

```js
const hasMetStreak = (questPlayStreakData?.current_playstreak_in_days ?? 0) >= (questMeta?.eligibility?.play_streak?.required_playstreak_in_days ?? Infinity);
```

eliobricenov (approved):

src/frontend/components/UI/QuestsViewer/components/QuestDetailsWrapper/index.tsx
BrettCleary: yeah I can change it to that. it will handle the case where `current_playstreak_in_days` and `required_playstreak_in_days` are null. That case shouldn't happen in practice but better to be safe

nyghtstalker (approved):
Confirmed that the "Claim all" button is now displayed instead of the "Sync" button and I was able to claim the credits.  

image
image

PR #1042, opened 2024-08-27
[Feat] add import game folder

BrettCleary:
# Summary
This PR adds the ability to import a game folder.

This could be useful if the user moved the game folder or deleted their hyperplay config folder.

# Testing 
https://app.qase.io/project/HC?suite=73&previewMode=modal&case=364


eliobricenov (approved):

red-game-dev (approved):

src/backend/ipcHandlers/importGameFolder.ts
flavioislima, line 1: One small improvement would be to check if a game is already in the library and is installed and skip it if yes.

src/backend/ipcHandlers/importGameFolder.ts
```diff
@@ -0,0 +1,49 @@
...
+        if (file_i.startsWith('0x') && file_i.endsWith('.json')) {
+          logDebug(`Found manifest file ${filePath_i}`, LogPrefix.HyperPlay)
+          const manifest = JSON.parse(readFileSync(filePath_i).toString())
+          if (
+            manifest &&
+            Object.hasOwn(manifest, 'manifest') &&
+            Object.hasOwn(manifest.manifest, 'appName')
+          ) {
+            const appName = manifest.manifest.appName
+            addGameToLibrary(appName)
```
flavioislima, line 38: This should be awaited or added under a try/catch because, since it does a request to our API, due to a race condition the import bellow could fail. 
https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/blob/432d021a6da358b664162bd14059fc3a14a4742f/src/backend/storeManagers/hyperplay/library.ts#L42

src/backend/ipcHandlers/importGameFolder.ts
BrettCleary: Hmm that is interesting. Maybe it would be good to check which version is newest (installed vs import candidate) and then to import if the import candidate is newer and keep the installed if that one is newer. But we don't enforce semver on the version string, so I don't think it would be possible to compare the versions 🤔

BrettCleary: So it is one or the other edge case assuming we cannot compare version strings

If we choose the import candidate, then the user might be moving from a newer version (installed) to an older version or vice versa. newer to older being the edge case

If we choose the installed, then maybe the user won't be able to import the newer version.

in the first case, they could delete the folder from the folder to be imported or import the other installed game again for that game

In the second case, they could uninstall first, then import.

Either way, this seems like a small edge case so I'd prefer to leave it as is, where it chooses the install candidate, if that's alright.

src/frontend/components/UI/ImportGameFolder/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+import styles from './index.module.scss'
+
+export function ImportGameFolder() {
+  const { t } = useTranslation()
+  const [gameFolderToImport, setGameFolderToImport] = useState('')
+
+  const importGameFolderMutation = useMutation({
+    mutationFn: async (gameFolder: string) => {
+      const result = await window.api.importGameFolder(gameFolder)
+      return result
```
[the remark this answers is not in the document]
BrettCleary: Great point. I added a success message now that importGameFolder returns. 

Since this is in advanced settings though, I'd like to handle further UI/UX improvements as needed on separate branches in the future. We would likely need some design input as well and I'm not sure if it's high priority at the moment.

src/backend/ipcHandlers/importGameFolder.ts
BrettCleary: fixed!

src/backend/ipcHandlers/importGameFolder.ts
```diff
@@ -0,0 +1,49 @@
...
+            importGame(appName, projectFolder_i)
```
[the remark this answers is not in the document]
BrettCleary: it will return on this api handler with the error message, which we then show under in the import button. It also gets logged to the console. I will change this to log to the console on the frontend though and to show a more user friendly error message.

Still a good idea to add try catch to the inner block though. I'll add one for the `file_i` import so that one broken import doesn't prevent the others from being imported.

src/backend/ipcHandlers/importGameFolder.ts
flavioislima: IMO this feature is to be used in pretty advanced cases and we can assume that it will be used more when moving installation through different computers. So I assume that most cases the person won't have other games installed on the imported game folder.

Checking only if the game is installed is fine, checking versions adds more complexity and to deal with a pretty small edge case so I agree to skip that and only check for the installed status.

Also, even if the person imports an older version, worst case scenario an update will be required and soon we will have a better update system in place so will be fine.

flavioislima:
Just need some testing steps now @BrettCleary

BrettCleary:
> Just need some testing steps now @BrettCleary

Thank you for the reminder. Good to go now @nyghtstalker

nyghtstalker (approved):
This works great! Love having this feature!

PR #1046, opened 2024-08-27
[TECH] Logs Cleanup

flavioislima:
- Changed the `stopChildProcesses` method to only log on Abort and for games only since it was logging all commands before.
- Removed most of other useless logs like the ones that says your are logged in or not on GOG and EPIC.
- Removed others that were added when some features were new and we were debugging like from abort controller and more.
- Didn't remove any HyperPlay logs.

### How to test

This should be easy to test and should not break anything since it is just logs removal. But since there were some changes to the `stopChildProcesses` method, would be good to test it again to see if games like XOciety will still close completely.

To test the removed logs do some actions like refresh the libraries, launch a game, exit, launch again, abort it from the overlay or the STOP button and check the HyperPlay logs (not game logs since they are not affected).

Logs including `PID` and `abortController` should not appear when refreshing library or closing the game normally. But the `Killing child process` ones should appear when the game is aborted from the client.

---


src/backend/launcher.ts
```diff
@@ -1020,7 +1017,10 @@ async function callRunner(
   return promise
 }
 
-async function stopChildProcesses(childPid: number) {
+async function stopChildProcesses(
+  childPid: number,
+  shouldLog: boolean = false
+) {
```
BrettCleary, line 1023: nitpick. could we make these params named and de-structure them here?

src/backend/launcher.ts
```diff
@@ -1065,24 +1067,30 @@ async function stopChildProcesses(childPid: number) {
...
       }
 
       stopProcessTree(childPid)
     } catch (error) {
-      logError(
-        `Error executing PowerShell command: ${error}`,
-        LogPrefix.Backend
-      )
+      if (shouldLog) {
+        logError(
```
BrettCleary, line 1089: Should we just make these `logDebug`? this will still be pretty spammy when a game is closed and I'm not sure it's necessary

BrettCleary (approved):

src/backend/launcher.ts
flavioislima: Yeah, will do.
they should appear only when aborting from the overlay or the Stop button.

red-game-dev (approved):

nyghtstalker (commented):
Confirmed that I am still able to force exit a game that didn't before (Bushi). Confirmed I am not seeing any `abortController` logs.

I am still seeing some `WARNING` with `PID` but not sure if these should be removed or not. 

image
image

nyghtstalker (approved):
Confirmed with Flavio that these PIDs are expected. Everything else looks good. Approved!

PR #1051, opened 2024-08-31
[Feat] One-click Install Steam on macOS

flavioislima:
Closes https://github.com/HyperPlay-Gaming/product-management/issues/429

## macOS Only
This adds a new Button besides the Add Game Button that Downloads and Install Steam for Windows using the compatibility layer on macOS.

### How to test
QASE test: https://app.qase.io/project/HC?case=362&suite=51&previewMode=modal

One thing to keep in mind is that the first launch of Steam takes a while since it needs to download and compille itself. The subsequent launches are faster. the first one took around 1 minute on my macbook.

---


src/backend/tools.ts
```diff
@@ -577,3 +588,122 @@ function getDxvkUrl(): string {
...
+ * Download and Install the Windows version of Steam on a Wine Prefix on macOS only
+ */
+export const SteamWindows = {
+  downloadSteam: async () => {
+    if (!isMac) {
+      return
+    }
+
+    const steamURL =
+      'https://cdn.cloudflare.steamstatic.com/client/installer/SteamSetup.exe'
```
jiyuu-jin, line 602: It's dope that this is just a static link!

jiyuu-jin (approved):
Code, LGTM, will test on mac today! 🚀

src/frontend/screens/Library/components/SteamInstall/SteamInstallDialog.tsx
```diff
@@ -0,0 +1,104 @@
...
+      setShowAlert('success')
+      setTimeout(() => setShowAlert('none'), 5000)
+      setIsInstalling(false)
+      onClose()
+    } catch (error) {
+      setShowAlert('danger')
+      setTimeout(() => setShowAlert('none'), 5000)
+      setIsInstalling(false)
+    }
+  }
```
eliobricenov, line 36: as a rule of thumb, whenever you need to have a loading state for an action, you're better using a mutation:

```tsx
const installSteamMutation = useMutation(
    () => window.api.installSteamWindows(),
    {
      onMutate: () => {
        setIsInstalling(true);
      },
      onSuccess: () => {
        setShowAlert('success');
        setTimeout(() => setShowAlert('none'), 5000);
        setIsInstalling(false);
        onClose();
      },
      onError: () => {
        setShowAlert('danger');
        setTimeout(() => setShowAlert('none'), 5000);
        setIsInstalling(false);
      },
      onSettled: () => {
        setIsInstalling(false);
      },
    }
  );

  const handleInstallSteam = () => {
    installSteamMutation.mutate();
  };
```

```tsx
<button 
    onClick={handleInstallSteam} 
    disabled={installSteamMutation.isLoading}
>
    {installSteamMutation.isLoading ? 'Installing...' : 'Install Steam'}
</button>
```

src/frontend/screens/Library/components/SteamInstall/SteamInstallButton.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const { t } = useTranslation()
+  useEffect(() => {
+    const checkWine = async () => {
+      const winelist: WineInstallation[] = await window.api.getAlternativeWine()
+      if (winelist.length > 0) {
+        setIsCompatibilityLayerAvailable(true)
+      }
+    }
+    checkWine()
+  }, [])
```
eliobricenov, line 49: I think this is fine, but if you need to check the loading status of this you could do:

```tsx
const { data: winelist, isLoading, isError } = useQuery<WineInstallation[]>(
    'alternativeWine',
    async () => {
      return await window.api.getAlternativeWine();
    }
  );

  const isCompatibilityLayerAvailable = winelist && winelist.length > 0;
```

src/frontend/screens/Library/components/SteamInstall/SteamInstallButton.tsx
```diff
@@ -0,0 +1,154 @@
...
+          {renderButtonText()}
+        </Button>
+      </Tooltip>
+      {showInstallDialog ? (
+        <SteamInstallDialog
+          isInstalling={isInstalling}
+          setIsInstalling={setIsInstalling}
+          onClose={() => setShowInstallDialog(false)}
+          setShowAlert={setShowAlert}
+        />
```
eliobricenov, line 122: I think the implementation should be:

```tsx
<SteamInstallDialog
    isInstalling={isInstalling}
    onInstall={installSteamMutation.mutate}
    onClose={() => setShowInstallDialog(false)}
/>
```

you can do the alert if you move the mutation to this component.

eliobricenov (dismissed):

src/frontend/screens/Library/components/SteamInstall/SteamInstallButton.tsx
eliobricenov: no need to pass the state setters

src/frontend/screens/Library/components/SteamInstall/SteamInstallDialog.tsx
flavioislima: I think I would need more changes than that though:
```Type '() => Promise<void>' has no properties in common with type 'UseMutationOptions<unknown, Error, void, unknown>'.ts(2559)```

the only improvement is that I won't need to pass the setState?  because it seems more code and the actual code is working fine afaik.

File: src/frontend/screens/Library/components/SteamInstall/SteamInstallButton.tsx (hunk 3 above)
Comment (flavioislima) [reply src/frontend/screens/Library/components/SteamInstall/SteamInstallButton.tsx:49]: I have some TS issues with this code so probably more changes than that will be necessary:

**Comment by flavioislima** [comment]
@eliobricenov implemented your suggested changes in a bit different way but still using the same logic.

File: src/frontend/screens/Library/components/SteamInstall/SteamInstallButton.tsx [hunk 5]
```diff
@@ -0,0 +1,179 @@
...
+      setIsInstalling(false)
+    },
+    onError: () => {
+      setShowAlert('danger')
+      setTimeout(() => setShowAlert('none'), 5000)
+      setIsInstalling(false)
+    },
+    onMutate: () => setIsInstalling(true),
+    mutationFn: async () => window.api.installSteamWindows()
+  })
```
eliobricenov, line 70: you don't need isInstalling anymore, you can just do `installSteamMutation.isLoading` it is doing the same.

eliobricenov (commented):
lgtm, left only one comment

eliobricenov (approved):

src/frontend/screens/Library/components/SteamInstall/SteamInstallButton.tsx
flavioislima: There is not `isLoading` prop though.
It is only status.

Also, if I dont use the isInstalling it doesnt change the state after finishing installing. Same for lunching.

nyghtstalker:
1. Unless this is intentional there is inconsistent punctuation at the end of each sentence. It should be the same for each sentence.

image

2. If cancelation is not an option once it is started should we hide this grayed out button?

image


Other than those two it works great! Super simple, smooth, and easy! Great job! I love it!!! 

image

flavioislima:
Nice!
I will fix the ponctuation and hide the cancel button. thanks for testing 🤘

nyghtstalker:
> Nice! I will fix the ponctuation and hide the cancel button. thanks for testing 🤘

Please let me know once that is complete so I can recheck and approve it.

flavioislima:
@nyghtstalker Did the changes, take a look when you have the time.

nyghtstalker (approved):
Punctuation has been fixed so they all end the same and button is hidden. Great job! 

image

PR #1052, opened 2024-09-01
[FIX] HyperPlay fails if the game default channel is not expected

flavioislima:
Should fix an issue where the Game fails to download or is stuck in the queue because the default channel is different from  `main`. 
So it was defaulting to `main` on `addDMToQueue` and then `release_meta` was coming as `undefined`.

Tested installing games that has only one main channel like Darkthrone, games that has two channels like Bushi and Code: Havoc that has only the Alpha channel and no main channel.

---

### DIFF
--- a/src/backend/downloadmanager/downloadqueue.ts
+++ b/src/backend/downloadmanager/downloadqueue.ts
@@ -133,7 +133,11 @@ async function addToQueue(element: DMQueueElement) {
   } else {
     const installInfo = await libraryManagerMap[
       element.params.runner
-    ].getInstallInfo(element.params.appName, element.params.platformToInstall)
+    ].getInstallInfo(
+      element.params.appName,
+      element.params.platformToInstall,
+      element.params.channelName ?? 'main'
+    )
 
     element.params.size = installInfo?.manifest?.download_size
       ? getFileSize(installInfo?.manifest?.download_size)
--- a/src/backend/storeManagers/hyperplay/library.ts
+++ b/src/backend/storeManagers/hyperplay/library.ts
@@ -16,7 +16,7 @@ import {
   sanitizeVersion,
   refreshGameInfoFromHpRelease
 } from './utils'
-import { getGameInfo as getGamesGameInfo } from './games'
+import { getGameInfo } from './games'
 import { getValistListingApiUrl, qaToken } from 'backend/constants'
 
 export async function addGameToLibrary(projectId: string) {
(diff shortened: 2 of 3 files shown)


jiyuu-jin (approved):
Sick 🤘 , you're a boss, jumping on this.

src/backend/downloadmanager/downloadqueue.ts
```diff
@@ -133,7 +133,11 @@ async function addToQueue(element: DMQueueElement) {
   } else {
     const installInfo = await libraryManagerMap[
       element.params.runner
-    ].getInstallInfo(element.params.appName, element.params.platformToInstall)
+    ].getInstallInfo(
+      element.params.appName,
+      element.params.platformToInstall,
+      element.params.channelName ?? 'main'
```
BrettCleary, line 139: do we still want the `?? 'main'` here? I don't see how this is different than `channelNameToInstall = 'main'` in the hp library

BrettCleary (approved):
lgtm

BrettCleary:
Is there a way that @nyghtstalker can test?

flavioislima:
> Is there a way that @nyghtstalker can test?

Yes, I will PM him a game that requires access code so he can test.

src/backend/downloadmanager/downloadqueue.ts
flavioislima: Not really, because in this method from `downloadqueue` it never comes as `undefined` since we always pass the channel to the `install` command and this ` ?? ` was just to satisfy typescript.

But not always when calling `getInstallInfo` it gets the channel and sometimes it gets `undefined`. 
I think it might be from a call somewhere else or a race condition where it miss the channel the first time and the second goes fine.

BrettCleary: are any error messages logged when it comes back `undefined`? @flavioislima

nyghtstalker (approved):
I tested installing the below and all install and launch without issues. This looks good to me!

1. Code Havoc (JOE CODE)
2. Dark Throne (Windows)
3. Dark Throne (Browser)
4. Bushi (Main)
5. Bushi (beta - JOE CODE)
6. Voxie Tactics

- Testing Code Havoc satisfies install without a MAIN channel
- Testing Bushi on both branches satisfies install with multiple channels
- Testing Dark Throne satisfies install with only one channel
- Testing Voxie Tactics satisfies install without specific channel

src/backend/downloadmanager/downloadqueue.ts
flavioislima: Only that release meta is undefined as a result. Since it gets nothing from an undefined channel.
That's is why the return.

PR #1059, opened 2024-09-06
fix: Preserve library status sort (#892)

### LINKED ISSUE #892: Library status sort is not retained when returning from a game page
### Describe the bug

When a user applies a status sort in the library, clicks into a game, and then returns the sort is removed.

### Add logs

N/A

### Steps to reproduce

1. Go to library
2. Select the "Alphabetical A-Z" status sort option
3. Click into any game
4. Use the Back arrow to return to the library
5. Observe that the status sort has been removed 

### Expected behavior

The status sort should be retained

### Screenshots

image

image

### hyperplay Version

Latest Stable

### System Information

- Windows 10
- hyperplay Version 0.13.1

### Additional information

_No response_

iratinho:
This fixes #892 where the library sort status is not being preserved when navigating to other pages.

### DIFF
--- a/src/frontend/screens/Library/index.tsx
+++ b/src/frontend/screens/Library/index.tsx
@@ -93,7 +93,7 @@ export default observer(function Library(): JSX.Element {
   // Track the screen view once and only once.
   useEffect(() => {
     window.api.trackScreen('Library')
-    libraryState.selectedFilter = filters[0]
+    libraryState.selectedFilter ??= filters[0]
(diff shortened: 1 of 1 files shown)


iratinho:
The build fails because it can't get the secrets for the authToken. This is actually a github protection, since this PR was created from a forked repository the secrets wont be available. This prevents malicious actors to try and dump all the secrets when running workflows by opening a PR.

PR #1063, opened 2024-09-10
[UI] Make Download Toast Draggable

flavioislima:
Change the behaviour of the `DownloadToastManager` to be draggable across the screen so it won't be on top of any other element.

Recordings:
Linux:
https://github.com/user-attachments/assets/c647834e-6c5b-4f71-b945-7b046d9f9fe7

macOS:
https://github.com/user-attachments/assets/79559724-bada-4296-aef3-2ea86c06a908

---


jiyuu-jin (approved):
Code LGTM, cool low hanging fruit, adds some nice functionality by just adding the draggable component.

eliobricenov (approved):

flavioislima:
> Code LGTM, cool low hanging fruit, adds some nice functionality by just adding the draggable component.

yes, this library is a beast. Pretty simple to use it anywhere in the code.

BrettCleary (approved):
Nice and simple! It works, but is there a reason why it doesn't follow the mouse cursor the same distance?

When I click to drag it to the top left, the element only moves ~0.6 pixels for every pixel my mouse cursor moves, which feels a little janky

flavioislima:
> Nice and simple! It works, but is there a reason why it doesn't follow the mouse cursor the same distance?
> 
> When I click to drag it to the top left, the element only moves ~0.6 pixels for every pixel my mouse cursor moves, which feels a little janky

Can you send a video?
I am using the defaults for this element, perhaps we can do some adjustments by passing some props to it.

BrettCleary:
> > Nice and simple! It works, but is there a reason why it doesn't follow the mouse cursor the same distance?
> > When I click to drag it to the top left, the element only moves ~0.6 pixels for every pixel my mouse cursor moves, which feels a little janky
> 
> For me its is working fine as showing in the videos above. Can you send a screen recording? I am using the defaults for this element, perhaps we can do some adjustments by passing some props to it.

after some more investigation, this only happens with the devtools console open. with it closed, it works as expected 👌

nyghtstalker (approved):
I love it! Super fun and cool! You can drag it to any page/screen and minimize it! Well done!

image
image

PR #1073, opened 2024-09-17
[FIX] Don't update games with removed platform build

flavioislima:
Fix issues like we had now with The Unfettered Awakening where players got the update available notification but the original build was removed only leaving the web one.

### How to Test

QASE Test: https://app.qase.io/project/HC?suite=39&previewMode=modal&case=366

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hyperplay",
-  "version": "0.18.2",
+  "version": "0.19.0",
   "private": true,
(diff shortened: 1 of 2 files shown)


BrettCleary (approved):
LGTM! a good check to add

nyghtstalker (approved):
Confirmed that when XOCIETY is set to an old version and `web` it is now displaying Play instead of Update!

image

PR #1076, opened 2024-09-18
[Fix] install modal folder

BrettCleary:
# Summary

This PR
- fixes a regression that did not allow the folder icon to be clickable on the download dialog so the path could not be changed
- Adds top margin to dialog footer so there is spacing between it and access codes 

image


src/frontend/components/UI/TextInputField/index.tsx
```diff
@@ -64,6 +64,7 @@ const TextInputField = ({
         )}
         rightSection={inputIcon}
         label={label}
+        readOnly={true}
```
eliobricenov, line 67: can we expose the props and use `readOnly={true}` where we need it instead of setting it for all implementations?

src/frontend/screens/Library/components/InstallModal/DownloadDialog/index.tsx
```diff
@@ -568,7 +568,7 @@ export default function DownloadDialog({
             value={installPath.replaceAll("'", '')}
             onChange={(event) => setInstallPath(event.target.value)}
             icon={<FontAwesomeIcon icon={faFolderOpen} />}
-            inputProps={{ readOnly: true, className: styles.folderTextInput }}
+            inputProps={{ className: styles.folderTextInput }}
```
eliobricenov, line 571: see other comment

eliobricenov (changes requested):

src/frontend/components/UI/TextInputField/index.tsx
BrettCleary: 🫡

BrettCleary: made the change. great suggestion

eliobricenov (approved):

BrettCleary:
> I cannot replicate the issue neither in `main` or on this branch. Is this windows only? tested on linux.


I don't see why it would be, but only tested on windows. just double checked on main and the folder icon is still unclickable

flavioislima (approved):
OK tested on macOS and Windows and could repliate the issue. On this branch it works fine.

nyghtstalker (approved):
Confirmed that the folder icon can be selected now and the added footer below the access codes.

image

PR #1078, opened 2024-09-19
[CI] Add i18n Parser Check

flavioislima:
The workflow check will fail if we forget to run `yarn i18n` when adding new translations.

---


flavioislima:
For testing, just check the first commit, it failed since there were some keys that were added but we forgot to update.
The second commit contains the keys and then the check passes.

.github/workflows/translations_parser.yml
BrettCleary, line 1: can we use our reusable workflows so this is easier to maintain?

BrettCleary: here is an example https://github.com/HyperPlay-Gaming/hyperplay-ui/blob/main/.github/workflows/lint.yml

BrettCleary (changes requested):

.github/workflows/translations_parser.yml
flavioislima: Tried but for some reason the parser fails everytime even though no keys were added to it.

BrettCleary: I think it needs to be `checkout_recursive: false` because if it is true you need to pass the `pat` secret. We don't need recursive checkout any more though since there are no submodules

BrettCleary: oh I see you tried that. taking a look at the latest

flavioislima: working now

BrettCleary (approved):
LGTM!

eliobricenov (approved):

PR #1079, opened 2024-09-19
fix(overlay): update layout to improve UX on lower viewports

eliobricenov:

### Summary of changes:
- made MM window non-floating
- added max-widths from https://www.figma.com/design/GKPELZxk551oVC3EDLgEDA/Current-Design?node-id=14888-158274&node-type=section&m=dev
- remove quest description unselect
- open the first quest description by default

---


flavioislima:
@eliobricenov can you add screenshots, how to test and which screen resolutions have you tested?

flavioislima (approved):

eliobricenov:
> @eliobricenov can you add screenshots, how to test and which screen resolutions have you tested?

Tested all screens down to steam-deck resolution (1300px)
image

src/frontend/components/UI/QuestsViewer/index.module.scss
```diff
@@ -18,6 +12,14 @@
...
+.detailsWrapper {
+  width: 1045px;
+  min-width: 800px;
+  max-width: 1063px !important;
+
+  @media (max-width: 1280px) {
+    width: 540px;
+    min-width: 520px;
+    max-width: 1045px !important;
+  }
```
eliobricenov, line 24: For some reason, I couldn't use the break-points mixins, every time I wanted to import the `_variables.scss` I would get this error:

```
[plugin:vite:css] [sass] Undefined mixin.
```

even though we should be importing it with the `import '@hyperplay/ui/style.css'` we have in `index.tsx`

src/frontend/components/UI/QuestsViewer/index.module.scss
```diff
@@ -18,6 +12,14 @@
...
+.detailsWrapper {
+  width: 1045px;
+  min-width: 800px;
+  max-width: 1063px !important;
+
+  @media (max-width: 1300px) {
+    width: 540px;
+    min-width: 520px;
+    max-width: 1045px !important;
+  }
```
eliobricenov, line 24: For some reason, I couldn't use the break-points mixins, every time I wanted to import the `_variables.scss` I would get this error:

```
[plugin:vite:css] [sass] Undefined mixin.
```

even though we should be importing it with the `import '@hyperplay/ui/style.css'` we have in `index.tsx`

jiyuu-jin (approved):
Code, LGTM

src/frontend/ExtensionManager/components/ExtensionContents/index.tsx
```diff
@@ -22,36 +24,47 @@ const animation = {
...
+  return (
+    <webview
+      nodeintegrationinsubframes="true"
+      webpreferences="contextIsolation=true, nodeIntegration=true"
+      src={`chrome-extension://${extensionState.extensionId}/notification.html`}
+      className={ExtensionContentsStyles.mmWindow}
+      allowpopups={trueAsStr}
+    ></webview>
+  )
+}
```
BrettCleary, line 49: could maybe combine these and pass in the `notification.html` or `popup.html` to keep it dry

BrettCleary (approved):
looks great

package.json
BrettCleary, line 1: if you want to hotfix, go ahead and bump the version to 0.19.2

src/frontend/ExtensionManager/components/ExtensionContents/index.tsx
eliobricenov: I think the code duplication is fine in this case

nyghtstalker (approved):
Confirmed that when the screen is reduced MM does not block the claim button.

image

PR #1054, opened 2024-09-04
[FIX] Don't show Game Settings for Browser games on game card

flavioislima:
Small fix to a bug I just found where we can access the game settings for Browser games.

Before:
image

After:
image

---

Use the following Checklist if you have changed something on the Backend or Frontend:

- [x] Tested the feature and it's working on a current and clean install.
- [ ] Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)
- [ ] Created / Updated Tests (If necessary)
- [ ] Created / Updated documentation (If necessary)

PR #1086, opened 2024-09-24
[FEAT] Add support for Actions when navigating to GamePage

flavioislima:
This is a small but useful change that should fix some UX issues.

It basically adds the ability to pass an action on the location state when navigating to the game page from anywhere, including the store.

The actions can be anything but so far I limited to `install`, `launch` and `uninstall`.

The api that navigates from the store was also update with that so we can either launch or install a game with one click instead of two.

Same for the download manager, the Play icon should call the Play method and the Download icon should show the Install dialog.
If the play icon was clicked but the game is not installed then it will also show the install dialog.

Merge this one before this one: https://github.com/HyperPlay-Gaming/hyperplay-store/pull/217

### How to test

- Launch/Add any game from the store.
- If the game is installed, it will launch it.
- If the game is not installed, will open the install dialog.
- From the Download manager screen
- The play button should launch the game if installed.
- The download button should open the install dialog.

---


src/backend/hyperplay_store_preload.ts
```diff
@@ -18,6 +19,7 @@ contextBridge.exposeInMainWorld('api', {
   apiVersion: 1,
   appIsInLibrary: async (gameId: string) =>
     ipcRenderer.invoke('appIsInLibrary', gameId, 'hyperplay'),
-  goToGamePage: (gameId: string) => ipcRenderer.send('goToGamePage', gameId),
+  goToGamePage: (args: { gameId: string; action: GamePageActions }) =>
```
BrettCleary, line 22:
```suggestion
  goToGamePage: (gameId: string; action?: GamePageActions ) =>
```

if we define it like this, we don't need to worry about backwards compatibility of the store with old versions of the client

src/common/types.ts
```diff
@@ -993,3 +993,5 @@ export interface PointsCollection {
   symbol: string
   image: string
 }
+
+export type GamePageActions = 'install' | 'launch' | 'uninstall'
```
BrettCleary, line 997: a best practice would be to put this type in https://github.com/HyperPlay-Gaming/utils/tree/main/src to keep the code dry and share with the store or other repos that might need this type (e.g. summon web portal)

BrettCleary: apologies if I missed it, but I don't think we are using uninstall anywhere. Can we remove it if it is not implemented?

src/frontend/screens/DownloadManager/components/DownloadManagerItem/index.tsx
```diff
@@ -127,20 +127,21 @@ const DownloadManagerItem = observer(({ element, current, state }: Props) => {
   const canceled = status === 'error' || (status === 'abort' && !current)
   const isExtracting = gameProgressStatus === 'extracting'
 
-  const goToGamePage = () => {
+  const goToGamePage = (action?: 'launch' | 'install') => {
```
BrettCleary, line 130:
```suggestion
  const goToGamePage = (action?: GamePageActions) => {
```

Can we type this?

src/frontend/screens/DownloadManager/components/DownloadManagerItem/index.tsx
BrettCleary, line 1: Do we have to handle launch and install on the frontend? My gut intuition is that this type of business logic should be handled in the main process

BrettCleary: I understand this would be a bigger refactor than can be handled on this PR due to the install modal popup and what not but is this a longer term goal?

src/frontend/screens/Game/GamePage/index.tsx
```diff
@@ -142,6 +144,7 @@ export default observer(function GamePage(): JSX.Element | null {
   const notSupportedGame =
     gameInfo.runner !== 'sideload' && gameInfo.thirdPartyManagedApp === 'Origin'
   const isOffline = connectivity.status !== 'online'
+  const isBrowserGame = Boolean(gameInfo.browserUrl)
```
BrettCleary, line 147:
```suggestion
    const isBrowserGame =
    installPlatform === 'Browser' || installPlatform === 'web'
```

^ copied from game card https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/blob/d8a155da6e66957c6051a796ab6a05cbd157c47c/src/frontend/screens/Library/components/GameCard/index.tsx#L167

I believe we just had an issue recently with this flag since some game infos have a browserUrl but are not installed browser games. I believe this happens when a game has a native and browser platform but the native platform is installed. This would still return true.

BrettCleary: yes you made that change on this pr 3 weeks ago https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/pull/1054

src/frontend/screens/Game/GamePage/index.tsx
```diff
@@ -153,6 +156,23 @@ export default observer(function GamePage(): JSX.Element | null {
     gameInstallInfo?.manifest?.download_size || 0
   )
 
+  const hasRun = useRef(false)
+  useEffect(() => {
+    if (!action || hasRun.current) return
+    hasRun.current = true
```
BrettCleary, line 162: Do we need the `hasRun` var? Looks like an anti-pattern

BrettCleary (changes requested):

src/backend/hyperplay_store_preload.ts
flavioislima: What if I told you that I wrote like this at first and then I thought that on the code reviews you guys would ask me to change to use the object instead? 🥲🤣🤣🤣🤣

Ok, will update those.

src/common/types.ts
flavioislima: Yeah, I thought on having it in case we want to implement uninstall from the store game page in the future. But I will remove it for now.

flavioislima: Thanks for reminding me of the utlils lib, I added the type there and updated both prs.

src/frontend/screens/Game/GamePage/index.tsx
flavioislima: Yeah, was to fix that issue with the game card.
I think the problem here is that sometimes we use 'Browser' and sometimes 'web'  and now 'webgl' so this is getting confusing and needing this type of workaround. 
Ideally we should receive the `isBrowserGame`  (or build) from the API so we would use it for everything and we could handle what is really a browser game from there.

src/frontend/screens/Game/GamePage/index.tsx
flavioislima: Yes, we need this otherwise the game launches twice.
This is just to make for the `useEffect` do not run twice.

src/frontend/screens/DownloadManager/components/DownloadManagerItem/index.tsx
flavioislima: done

src/common/types.ts
flavioislima: done and done

src/backend/hyperplay_store_preload.ts
flavioislima: done

src/frontend/screens/Game/GamePage/index.tsx
BrettCleary: Well `webgl` isn't for a web game. It is just a one-off for the Unfettered team really. The api just supplies a map of platform keys to platform meta for a release and it's up to the client to decide which platform to install, whether that be browser or native. 

I think the suggested change above works `installPlatform === 'Browser' || installPlatform === 'web'`. I think `"Browser"` is just for backwards compatibility? `web` is the platform key for a browser game in a release currently

eliobricenov (approved):
lgtm

PR #1087, opened 2024-09-24
[FEAT] Client IPDT Patching implementation

flavioislima:
This implements the client side IPDT patching for HyperPlay games.
Details and how to test we should discuss internally since it is a lot of information to write on the PR.

---


BrettCleary:
looks like `any` ts errors are causing codecheck and test GHA's to fail. Could we fix the type or add a `// @ts-expect-error` before each line @flavioislima

flavioislima:
> looks like `any` ts errors are causing codecheck and test GHA's to fail. Could we fix the type or add a `// @ts-expect-error` before each line @flavioislima 

I think was a tsconfig I changed, will revert it.

src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -1475,3 +1561,345 @@ function writeManifestFile(
...
+          )}`,
+          LogPrefix.HyperPlay
+        )
+      }
+    }
+  } catch (error) {
+    logError(`Error while patching ${error}`, LogPrefix.HyperPlay)
+
+    trackEvent({
+      event: 'Patching Failed',
```
BrettCleary, line 1803: do we already have some charts for this? I'd like to set up alerting

src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -1475,3 +1561,345 @@ function writeManifestFile(
...
+    properties: {
+      game_name: gameInfo.app_name,
+      game_title: gameInfo.title,
+      platform: getPlatformName(platform),
+      platform_arch: platform
+    }
+  })
+
+  logInfo(`Patching ${appName} completed`, LogPrefix.HyperPlay)
+  return { status: 'done' }
```
BrettCleary, line 1843: couldn't this be moved into the try block above? It would be closer to the code's context there

src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -1475,3 +1561,345 @@ function writeManifestFile(
...
+      ? manifestPath.replace(/\\/g, '\\\\').replace(/ /g, '\\ ')
+      : manifestPath
+
+  if (!existsSync(normalizedManifestPath)) {
+    logDebug(
+      `Manifest for ${appName} not found for version ${version} and platform ${platformName}`,
+      LogPrefix.HyperPlay
+    )
+
+    try {
```
BrettCleary, line 1866: shouldn't we put all of this function's code in the try block?

src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -1394,17 +1421,59 @@ export async function update(
...
+      accessCode: args?.accessCode,
+      updateOnly: true,
+      siweValues: args?.siweValues
+    })
+    if (isMarketWars) {
+      try {
+        await runModPatcher(appName)
+      } catch (error) {
+        return { status: 'error' }
+      }
```
BrettCleary, line 1472: this type of code snippet is a result of us mixing JS errors and our own status string and having to convert between the two. 

In the future, I think we could simplify this considerably if we threw a custom error object with a status property. here's an example of how we could do that https://github.com/HyperPlay-Gaming/hyperplay-dev/blob/main/utils/EndpointError.ts

src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -1475,3 +1561,345 @@ function writeManifestFile(
...
+  signal: AbortSignal
+): Promise<InstallResult> {
+  const patcherisEnabled = ldMainClient.variation('enable-patcher', false)
+
+  if (!patcherisEnabled) {
+    return { status: 'error' }
+  }
+
+  const appName = gameInfo.app_name
+  const window = getMainWindow()
```
BrettCleary, line 1659: nitpick. could we rename this mainWindow?

BrettCleary (commented):
still going through the PR, but wanted to get these comments out. beast PR! 🔥

src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -1475,3 +1561,345 @@ function writeManifestFile(
 
   return store.set('manifest', installedInfo)
 }
+
+export const downloadPatcher = async () => {
+  if (!existsSync(ipdtPatcher)) {
```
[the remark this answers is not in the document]
flavioislima: Yeah, I am working on a solution to update it on every app update.

src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -1475,3 +1561,345 @@ function writeManifestFile(
...
+  }
+
+  const appName = gameInfo.app_name
+  const window = getMainWindow()
+  const { install_path, version, platform } = gameInfo.install
+  const { maxWorkers } = GlobalConfig.get().getSettings()
+  let aborted = false
+
+  if (!existsSync(ipdtPatcher)) {
+    await downloadPatcher()
```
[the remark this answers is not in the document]
flavioislima: Yeah, that makes sense.
But just to clarify, this method is only called here if it fails to download on the client start.

src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -1475,3 +1561,345 @@ function writeManifestFile(
...
+  logInfo(
+    `Patching ${gameInfo.title} from ${version} to ${newVersion}`,
+    LogPrefix.HyperPlay
+  )
+
+  try {
+    let totalBlocks = 0
+    let downloadedBlocks = 0
+    let downloadedData = 0
+    const blockSize = 512 * 1024 // 512KB in bytes
```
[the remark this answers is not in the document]
flavioislima: Yeah, was discussing this with Zach, the best way to fix this is to have the size of the block on the manifest, as well the total size of all blocks.
Right now the calculation is never 100% and sometimes the patching might finish before the 100% or, worse, keep going after 100%.
For MVP, I think this is fine but needs to be improved.

src/backend/storeManagers/hyperplay/games.ts
flavioislima: No, just added these events for now. Didn't have time to create the charts yet but should be straightforward.

src/backend/storeManagers/hyperplay/games.ts
flavioislima: It cannot go above the `if (aborted)`. I needed this check here to cover some edge cases on Windows.
Linux and macOS deal fine with the AbortController but Windows needs more checks otherwise it will keep going to the `done` status and the game will have the info updated without being updated.

BrettCleary: yeah i was thinking the `if (aborted)` would go in the try block too

src/backend/storeManagers/hyperplay/games.ts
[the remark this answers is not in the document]
BrettCleary: Yeah but if it fails on start, I'd argue that it is more likely to fail again here as well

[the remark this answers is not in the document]
BrettCleary: Either way if the ipdt download fails here, `applyPatching` shouldn't proceed

src/backend/storeManagers/hyperplay/games.ts
flavioislima: Ah I see now. Will do

src/backend/storeManagers/hyperplay/games.ts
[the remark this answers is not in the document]
flavioislima: Done

src/backend/main.ts
```diff
@@ -672,9 +672,21 @@ if (!gotTheLock) {
 
     initTrayIcon(mainWindow)
 
+    await HyperPlayGameManager.downloadPatcher()
+      .then(async () => {
+        logInfo('Patcher downloaded', LogPrefix.HyperPlay)
+      })
+      .catch((e) => {
+        logError(`Error downloading patcher: ${e}`, LogPrefix.HyperPlay)
+      })
```
BrettCleary, line 681: Could we refactor to try/catch here?

src/frontend/screens/DownloadManager/components/ProgressHeader/index.tsx
```diff
@@ -50,6 +51,11 @@ export default function ProgressHeader(props: {
   const showDownloadBar =
     props.state !== 'idle' && props.appName && progress.percent
 
+  const progressBytes =
+    typeof progress.bytes === 'number'
+      ? size(progress.bytes * 1024 * 1024)
+      : progress.bytes
+
```
eliobricenov, line 58: this seems off, if progress.bytes is typed as a string why is it possible it is a number?

src/backend/main.ts
```diff
@@ -920,7 +932,10 @@ ipcMain.on('navigate', async (event, appName) => {
 
 ipcMain.handle('getGameInfo', async (event, appName, runner) => {
   // Fastpath since we sometimes have to request info for a GOG game as Legendary because we don't know it's a GOG game yet
-  if (runner === 'legendary' && !LegendaryLibraryManager.hasGame(appName)) {
+  if (
+    !appName ||
```
eliobricenov, line 911: is this to catch empty strings?

src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -1475,3 +1558,360 @@ function writeManifestFile(
...
+    )
+      .then(async () => {
+        // remove the empty line at the end of the json file to avoid EOF issues
+        const manifestContent = await readFile(manifestPath, 'utf8')
+        const manifestContentTrimmed = manifestContent.trim()
+        await writeFile(manifestPath, manifestContentTrimmed)
+      })
+      .catch((error) => {
+        throw new Error(`Error downloading manifest for ${appName}: ${error}`)
+      })
```
eliobricenov, line 1637: this is already in a try-catch with the same error logging, why are we using a then.catch?

src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -1475,3 +1558,360 @@ function writeManifestFile(
...
+    if (!channelName || !platform || !is_installed) return false
+
+    // download only if the manifest file is not already downloaded and its the same version
+    const manifestName = `${appName}-${platform}-${version}.json`
+    const manifestPath = path.join(ipdtManifestsPath, manifestName)
+    if (existsSync(manifestPath)) return false
+
+    const releaseId = generateID(appName, version)
+    const manifestUrl = await getHyperPlayReleaseManifest(releaseId, platform)
+    if (!manifestUrl) return false
```
eliobricenov, line 1616: from what I see, the return value of this function is not being used anywhere. For errors why not throw them instead of silently returning a boolean?

src/backend/storeManagers/hyperplay/games.ts
flavioislima: Yeah, in my tests there was a race conditions that were fixed when I added this extra then/catch.

src/frontend/screens/DownloadManager/components/ProgressHeader/index.tsx
flavioislima: yeah, so, its complicated.
For Legendary and GOGDL it is a string, but for hyperplay its is a number.
We need to refactor those in the future to be the same for all.

src/backend/main.ts
flavioislima: exactly, we are getting a lot of `app not found in library` because it is coming as empty sometimes, we need to investigate this further and probably avoid the call if no appName on frontend.

src/backend/storeManagers/hyperplay/games.ts
flavioislima: yeah, true

flavioislima: Changed it now to throw errors and don't return anything else.

src/backend/main.ts
eliobricenov: let's add a TODO, and a logspecifying the early return

```ts
logWarn('skipping get game info because app name is empty')
```

src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -1394,17 +1417,60 @@ export async function update(
...
+      channelName: gameInfo.install.channelName,
+      accessCode: args?.accessCode,
+      updateOnly: true,
+      siweValues: args?.siweValues
+    })
+    if (isMarketWars) {
+      try {
+        await runModPatcher(appName)
+      } catch (error) {
+        return { status: 'error' }
```
eliobricenov, line 1468: let's add another log here

src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -1475,3 +1558,358 @@ function writeManifestFile(
...
+    const {
+      install: { channelName, platform },
+      is_installed
+    } = getGameInfo(appName)
+    if (!channelName || !platform || !is_installed) throw Error('Invalid game')
+
+    // download only if the manifest file is not already downloaded and its the same version
+    const manifestName = `${appName}-${platform}-${version}.json`
+    const manifestPath = path.join(ipdtManifestsPath, manifestName)
+    if (existsSync(manifestPath)) return
```
eliobricenov, line 1612: let's add a log here

src/backend/storeManagers/hyperplay/games.ts
eliobricenov: can you give it another try? the code should be he same the await syntax

src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -1475,3 +1558,358 @@ function writeManifestFile(
...
+      install_path,
+      currentManifest,
+      previousManifest,
+      ipfsGateway,
+      ipfsGateway,
+      signal
+    )
+
+    if (signal.aborted) {
+      return { status: 'abort' }
```
eliobricenov, line 1743: add log here

src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -1475,3 +1558,358 @@ function writeManifestFile(
...
+    signal.onabort = () => {
+      aborted = true
+      return { status: 'abort' }
+    }
+
+    for await (const output of generator) {
+      logInfo(output, LogPrefix.HyperPlay)
+
+      if (signal.aborted) {
+        return { status: 'abort' }
```
eliobricenov, line 1755: add log here

eliobricenov (commented):
I like your approach to logging, let's add more logs in the early returns to make sure we can debug errors

jiyuu-jin:

jiyuu-jin:
Recheck

src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -1475,3 +1559,351 @@ function writeManifestFile(
...
+
+  try {
+    const { patchFolder } = await import('@hyperplay/patcher')
+
+    const mainWindow = getMainWindow()
+    let aborted = false
+
+    if (!existsSync(ipdtPatcher)) {
+      await downloadPatcher()
+    }
```
BrettCleary, line 1673: I'd say let's download every time to get the latest version while this is feature flagged and open a separate PR to only download if there is a new version 

```suggestion
    await downloadPatcher()
```

BrettCleary: we'd have to remove this line in `downloadPatcher` too

```
if (!existsSync(ipdtPatcher)) {
```

flavioislima: ok, sounds good

src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -77,6 +91,11 @@ import getPartitionCookies from 'backend/utils/get_partition_cookies'
 import { prepareBaseGameForModding } from 'backend/ipcHandlers/mods'
 import { runWineCommandOnGame } from 'backend/utils/compatibility_layers'
 
+import { chmod, writeFile } from 'fs/promises'
+import { trackEvent } from 'backend/api/metrics'
```
BrettCleary, line 95:
```suggestion
import { trackEvent } from 'backend/metrics/metrics'
```

without this, we are attempting to send the event via the ipcRenderer and get an error message like this in the console

```
TypeError: Cannot read properties of undefined (reading 'invoke')
    at trackEvent 
    at applyPatching 
    at async Module.update$2 [as update] 
    at async updateQueueElement 
    at async initQueue 
```

src/backend/storeManagers/hyperplay/games.ts
flavioislima: solved

src/backend/storeManagers/hyperplay/games.ts
flavioislima: true, autoimports sometimes imports from the api calls. will fix

BrettCleary (changes requested):
getting close 🚀

src/backend/storeManagers/hyperplay/library.ts
```diff
@@ -236,6 +239,13 @@ export async function listUpdateableGames(): Promise<string[]> {
           val.channels[val.install.channelName].release_meta.platforms
         ).includes(val.install.platform)
       ) {
+        if (val.channels[val.install.channelName]) {
+          downloadGameIpdtManifest(val.app_name, val.install.version)
+          downloadGameIpdtManifest(
+            val.app_name,
+            val.channels[val.install.channelName].release_meta.name
+          )
```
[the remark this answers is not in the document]
flavioislima: It only downloads if the manifest doesn't exists already.
But yeah, at first I was only downloading when clicking update but then there was a racing condition there and the patching was failing. I will try something else to fix that and remove it from here.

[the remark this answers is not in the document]
flavioislima: Ok, moved back to only download before the updates. It is working fine after moving to `fetch`

src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -1475,3 +1560,389 @@ function writeManifestFile(
...
+    const blockSize = 512 * 1024 // 512KB in bytes
+    const startTime = Date.now()
+
+    if (!existsSync(datastoreDir)) {
+      mkdirSync(datastoreDir, { recursive: true })
+    }
+
+    const { maxWorkers } = GlobalConfig.get().getSettings()
+
+    console.log('patcher params---------', {
```
BrettCleary, line 1733: double checking we want to keep this in

BrettCleary (approved):
LGTM. we should do one final round of testing, merge, and handle further improvements on other branches

src/backend/storeManagers/hyperplay/games.ts
flavioislima: nope, removed it

src/backend/downloadmanager/downloadqueue.ts
```diff
@@ -139,9 +141,28 @@ async function addToQueue(element: DMQueueElement) {
...
+      const updateSize =
+        // @ts-expect-error - confusion because of the platform
+        element.params.gameInfo.channels[element.channel!].release_meta
+          .platforms[element.params.platformToInstall].downloadSize
+          ? getFileSize(
+              // @ts-expect-error - confusion because of the platform
+              element.params.gameInfo.channels[element.channel!].release_meta
+                .platforms[element.params.platformToInstall].downloadSize
+            )
+          : '?? MB'
```
eliobricenov, line 160: This ternary is really hard to read.

```ts
let updateSize = '?? MB'

if (element.channel && element.params.gameInfo.channels) {
  const releasePlatforms =
    element.params.gameInfo.channels[element.channel].release_meta
      .platforms

  const platformConfig =
    // @ts-expect-error - key types do not match the types but it is still a string map
    releasePlatforms[element.params.platformToInstall]

  if (platformConfig?.downloadSize) {
    updateSize = getFileSize(platformConfig.downloadSize)
  }
}
```

src/backend/storeManagers/hyperplay/games.ts
```diff
@@ -1475,3 +1564,387 @@ function writeManifestFile(
...
+      {
+        extra: {
+          method: 'getManifest',
+          appName,
+          platformName,
+          version
+        }
+      }
+    )
+    return ''
```
eliobricenov, line 1928: what's the benefit of returning an empty string as the manifest?

src/backend/downloadmanager/downloadqueue.ts
```diff
@@ -340,6 +361,28 @@ function processNotification(element: DMQueueElement, status: DMStatus) {
...
+  const queue: DMQueueElement[] = downloadManager.get('queue', [])
+
+  const index = queue.findIndex(
+    (el) => el.params.appName === element.params.appName
+  )
+  if (index !== -1) {
+    queue[index][prop] = value
+    if (currentElement) {
+      currentElement[prop] = value
+    }
```
eliobricenov, line 378: This is a side effect not really tied to what this function is for. What do you think about adding a `updateActiveQueueElement` function?

eliobricenov (approved):

src/backend/downloadmanager/downloadqueue.ts
flavioislima: Fixed

src/backend/storeManagers/hyperplay/games.ts
flavioislima: None, remove this line, it will never get there.

src/backend/downloadmanager/downloadqueue.ts
flavioislima: Done

nyghtstalker (approved):
### We changed the scope to **_only incorporate Windows for this release_**. This will not apply to Mac and Linux - those will be added at a later date.

Confirmed that patching is working correctly on Windows and the games are launching after being patched.

image
image

jiyuu-jin:
> ### We changed the scope to **_only incorporate Windows for this release_**. This will not apply to Mac and Linux - those will be added at a later date.
> Confirmed that patching is working correctly on Windows and the games are launching after being patched.
> 
> image image

Did the patch actually take an hour for you?

PR #1088, opened 2024-09-24
fix(overlay): set autoBind true to AuthState makeAutoObservable

eliobricenov:

### Summary

This was happening due to the context of `this` being lost in this function handler:

```jsx
<QuestDetailsWrapper
   ...
    openSignInModal={authState.openSignInModal}
/>
```

See https://stackoverflow.com/a/69100459 for reference.

### How to test:

On Production:
- While being logged out click the sign in button from the overlay.
image
- Check that nothing happens.

In the terminal, you'll see this error:
```
(14:29:02) ERROR:   [Frontend]:         TypeError: Cannot set properties of undefined (setting 'signInModalOpen')
    at openSignInModal (http://localhost:5173/src/frontend/state/authState.ts:27:26)
    at executeAction (http://localhost:5173/node_modules/.vite/deps/chunk-4TF65JMT.js?v=526fa203:1117:15)
    at openSignInModal (http://localhost:5173/node_modules/.vite/deps/chunk-4TF65JMT.js?v=526fa203:1102:12)
    at HTMLUnknownElement.callCallback2 (http://localhost:5173/node_modules/.vite/deps/chunk-I3COAS7K.js?v=526fa203:3674:22)
```

With these changes, the login modal should pop up.

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -169,7 +169,7 @@
     "@fortawesome/react-fontawesome": "^0.2.2",
     "@hyperplay/chains": "^0.3.0",
     "@hyperplay/check-disk-space": "^3.5.2",
-    "@hyperplay/quests-ui": "^0.0.20",
+    "@hyperplay/quests-ui": "^0.0.21",
     "@hyperplay/ui": "^1.7.18",
     "@hyperplay/utils": "^0.0.16",
     "@mantine/carousel": "^7.12.0",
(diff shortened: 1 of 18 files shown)


eliobricenov:
This technically would happen with all the stores, but we probably don't have another implementation like this. We need to be aware of this for the future.

BrettCleary (approved):
lgtm. yup I think elsewhere we use arrow functions but good thing to look out for like you said

nyghtstalker (approved):
Confirmed that the login modal appears after clicking the Sign In button on the quests and the user is able to successfully login from there.

image

PR #1094, opened 2024-09-26
fix(overlay): show notification window on transaction request

eliobricenov:

With the update to the overlay layout, there was a regression in the MM window, and now, when the overlay is open and there is a transaction request, we don't show the notification popup automatically, closing and re-opening the overlay is needed.

You can test this in dev mode in `main` by connecting your wallet and requesting a signature from the console:

```
await window.ethereum.request({
        method: 'personal_sign',
        params: ['Hi', (await window.ethereum.request({ method: 'eth_requestAccounts' }))[0]]
      });
 ```

---

### DIFF
--- a/src/frontend/ExtensionManager/components/ExtensionContents/index.tsx
+++ b/src/frontend/ExtensionManager/components/ExtensionContents/index.tsx
@@ -5,8 +5,6 @@ import ExtensionContentsStyles from './index.module.scss'
 
 const trueAsStr = 'false' as unknown as boolean | undefined
 
-type ExtensionState = 'popup' | 'notification'
-
 //Module type augmentation necessary to use experimental feature nodeintegrationinsubframes
 //https://www.electronjs.org/docs/latest/api/webview-tag
 /* eslint-disable react/no-unknown-property */
@@ -26,7 +24,8 @@ const animation = {
   transition: { duration: 0.2 }
 }
 
-export function ExtensionPopup({ state }: { state: ExtensionState }) {
+export const ExtensionContents = observer(() => {
+  const state = extensionState.isPopupOpen ? 'popup' : 'notification'
   return (
     <webview
       nodeintegrationinsubframes="true"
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/frontend/ExtensionManager/index.tsx [hunk 1]
```diff
@@ -42,7 +43,7 @@ const ExtensionManager = function () {
   }
 
   if (isOverlay) {
-    return <ExtensionPopup />
+    return <ExtensionContents />
```
eliobricenov, line 46: This was the problem, we were showing the normal MM popup regardless of the state in the ExtensionStore

src/frontend/ExtensionManager/components/ExtensionContents/index.tsx
```diff
@@ -24,47 +25,40 @@ const animation = {
...
-    <webview
-      nodeintegrationinsubframes="true"
-      webpreferences="contextIsolation=true, nodeIntegration=true"
-      src={`chrome-extension://${extensionState.extensionId}/notification.html`}
-      className={ExtensionContentsStyles.mmWindow}
-      allowpopups={trueAsStr}
-    ></webview>
-  )
+export function ExtensionContents() {
+  const state = extensionState.isPopupOpen ? 'popup' : 'notification'
```
BrettCleary, line 41: shouldn't this be wrapped in observer?

src/frontend/ExtensionManager/components/ExtensionContents/index.tsx
```diff
@@ -24,47 +25,40 @@ const animation = {
   transition: { duration: 0.2 }
 }
 
-export function ExtensionPopup() {
+export function ExtensionPopup({ state }: { state: ExtensionState }) {
```
BrettCleary, line 28: since this uses the mobx state, may want to wrap this in observer too even though extensionId won't change

eliobricenov: you're right, this is working because the OverlayManager is wrapped in an observer, but this component should be wrapped regardless.

jiyuu-jin (approved):
Makes sense, looks good.

BrettCleary (approved):
LGTM. let's be sure to test all the overlay flows in QA on this one

nyghtstalker (approved):
Checked all flows I could think of and E2E testing. Confirmed that the trx approval screen does appear after attempting to claim a reward.

PR #1066, opened 2024-09-11
[FIX] Wrong info on Download Manager

flavioislima:
Fixes:
- Wrong store name for HyperPlay Games
- Missing Install size on some games
- Missing Date of start and completition

Captura de imagem_20240911_125052
 

---

Use the following Checklist if you have changed something on the Backend or Frontend:

- [x] Tested the feature and it's working on a current and clean install.
- [ ] Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)
- [ ] Created / Updated Tests (If necessary)
- [ ] Created / Updated documentation (If necessary)

PR #1095, opened 2024-10-01
[FIX] Download Manager Toast Information

flavioislima:
This refactors part of the Download Manager Toast component so it can use the same progress that we use in other places like the game page and the download page.
This should fix the differences in downloaded bytes and ETA, especially for the extracting method that its not precise right now.

Also added a new Hook that we can use as a source of truth to get both the `gameInfo` and the `installInfo` from any game.

### Current issues
Captura de imagem_20241001_122405

Captura de imagem_20241001_122451

### Fixed on this branch
Captura de imagem_20241001_121056
(Note that the download size is 0 due to a bug fixed on #1066)

---


src/frontend/hooks/useGameInfo.ts
```diff
@@ -0,0 +1,56 @@
...
+      }
+
+      if (newInfo) {
+        setGameInfo(newInfo)
+      }
+    }
+    getNewInfo()
+  }, [appName, runner, platform, channel])
+
+  return { gameInfo, installInfo }
```
eliobricenov, line 55: This can be a query, you get pending states, caching, and error handling 🙏

```tsx
const query = useQuery({
    queryKey: ['gameData', appName, runner, platform, channel],
    queryFn: async () => {
      const gameInfo = await getGameInfo(appName, runner);
      
      let installInfo: InstallInfo = null;
      if (gameInfo && platform) {
        installInfo = await getInstallInfo(
          appName,
          gameInfo.runner,
          platform,
          channel
        );
      }

      return { gameInfo, installInfo };
    },
  });

  return {
    ...query.data,
    isLoading: query.isLoading,
    isError: query.isError,
    error: query.error,
  };
  ```

File: src/frontend/hooks/hasProgress.ts [hunk 2]
```diff
@@ -51,5 +61,23 @@ export const hasProgress = (appName: string) => {
     }
   }, [appName])
 
-  return [progress, previousProgress]
+  useEffect(() => {
+    if (isExtracting) {
+      setProgress(nullProgress) // reset progress to 0
+    }
+  }, [])
```
eliobricenov, line 68: I know this is legacy code but we don't need this useEffect. We can derive the value:

```tsx
return {
    progress: isExtracting ? nullProgress : progress,
    previousProgress,
    etaInMs: convertEtaToMs(progress.eta || '00:00:00')
}
```

eliobricenov: context: https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state

eliobricenov (commented):
overall lgtm, left a couple comments

src/frontend/hooks/hasProgress.ts
flavioislima: Cool, will check this out

src/frontend/hooks/useGameInfo.ts
flavioislima: I knew it!

src/frontend/hooks/hasProgress.ts
flavioislima: This makes the progress always 0 when extracting, so I think we need the use effect because then it resets only once.

src/frontend/hooks/useGameInfo.ts
flavioislima: done the change

eliobricenov (approved):

src/frontend/hooks/hasProgress.ts
eliobricenov: Oh, gotcha, we need to reset it once. I think that's fine, then, either way, it seems this pattern is the recommended https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes, but tbh I've never used it.

BrettCleary (approved):
LGTM

jiyuu-jin (approved):

nyghtstalker:
@flavioislima I confirmed I am seeing the download times on the library game page and download page match the download modal. 

- I confirmed on `main` and this is not occurring. My concern is this PR will bring this into `main`. The size is displaying as 0 on the Downloading and Downloaded tabs.

image
image
image
image

flavioislima:
> @flavioislima I confirmed I am seeing the download times on the library game page and download page match the download modal. 
> 
> - I confirmed on `main` and this is not occurring. My concern is this PR will bring this into `main`. The size is displaying as 0 on the Downloading and Downloaded tabs.
>
> 

That is because it did not was updated with main, I did it now if you want to test again.
I started this pr before the other was merged.

PR #1097, opened 2024-10-02
[Feat] Support quests for non-hp games

BrettCleary:
# Summary

Support quests for games in external game stores.

Open the overlay for these games if a quest exists for it, and redirect to the appropriate store url if the game is not in the user's library.

# Testing
Search for test Fortnite quest to test


eliobricenov (approved):

flavioislima (approved):

nyghtstalker:
The only quest/game I was able to test was Fortnite. I can confirm that I am seeing the overlay appear when launching the game.

1. Unable to claim the reward once the counter completes. After relaunch the button is clickable but gives an error.

2. Claimed page still shows the quest content on the right side. Normally that side is blank. I don't think this is a blocker but would be good to get this fixed if it is not intended.

image
image
image

PR #1100, opened 2024-10-03
feat(quest): add support for play streak progress sync

eliobricenov:

steps to test: https://app.qase.io/case/HC-367

---


src/frontend/ExtensionHandler/index.tsx
```diff
@@ -5,6 +5,7 @@ import { useNavigate } from 'react-router-dom'
 
 declare global {
   interface Window {
+    // @ts-expect-error: conflict with wagmi ethereum types
```
eliobricenov, line 8: If we don't do this, the global type definition of wagmi conflicts with this.

src/frontend/store/OnboardingStore.tsx
```diff
@@ -13,6 +13,9 @@ class OnboardingStore {
 
   constructor() {
     makeAutoObservable(this)
+    window.api.handleOpenOnboarding(() => {
+      this.openOnboarding()
+    })
```
eliobricenov, line 18: we need this to be able to trigger opening the modal in the main window from another.

src/frontend/components/UI/QuestsViewer/index.tsx
BrettCleary, line 1: Could we handle the case where the user clicks "Sync Progress" but is not logged in with an alert message inside the quest details component that shows after clicking sync progress? 

I know we have the alert above the quest components in the quests page and the overlay, but I think users could miss that.

BrettCleary: I think we should also consider only showing the sync progress button for games where the user has a sync-able quest. If the user isn't logged in in this case, then we wouldn't show the sync progress button so that would solve the above comment as well. 

And then we should hide the button after the user finishes the sync.

BrettCleary: we may want to consider also hiding it if the user already has a completed streak or else they may waste a sync if we want to continue to use this pattern in the future. I just checked and it lets you use your sync even if you've already completed the streak

image

src/frontend/screens/Quests/components/QuestDetailsViewPlay/index.tsx
```diff
@@ -12,222 +12,263 @@ import { useNavigate } from 'react-router-dom'
...
-          if (!res) {
-            throw new Error('Game not found in library')
+      if (questsWithExternalSync.includes(questMeta.id)) {
+        syncPlayStreakWithExternalSource(questMeta.id)
+      } else {
+        await window.api.syncPlaySession(questMeta.project_id, 'hyperplay')
+        await questPlayStreakResult.invalidateQuery()
+        window.api.logInfo(`Synced play session for quest ${questMeta.id}`)
+        console.log(`Synced play session for quest ${questMeta.id}`)
+      }
```
BrettCleary, line 66: we should not be syncing the playsession from the quests page since play streaks can only be earned by playing the game, which is why we sync in the overlay quest details

src/frontend/screens/Quests/components/QuestDetailsViewPlay/index.tsx
BrettCleary, line 1: If I click sync progress with the incorrect wallet connected, I do not get an error message or any visual change. I see "Playstreak synced with external source" logged to the console.

When I swap to the correct address, the signature request shows.

src/frontend/components/UI/QuestsViewer/index.tsx
BrettCleary: looks like we could easily hide if it is not in `flags.questsWithExternalSync` but might require more work to hide if the wrong wallet account is connected. we would need to listen to accountChanged events in the main process I think. It would solve the issue I pointed out in `src/frontend/screens/Quests/components/QuestDetailsViewPlay/index.tsx` in this review though.

BrettCleary (changes requested):
looking good. it's close

eliobricenov:

eliobricenov:
recheck

src/frontend/App.tsx
BrettCleary, line 1: this is not rendered in the overlay. check `src/frontend/ViewManager.tsx`. 

So I don't think this will be called. you probably want to move `AlertToast` to `src/frontend/OverlayManager/index.tsx` so it will show the alert in the overlay

src/frontend/hooks/useSyncPlayStreakWithExternalSource.ts
```diff
@@ -0,0 +1,104 @@
...
+      currentProvider === 'Unconnected' && !web3Provider
+
+    // we do window.api.focusMainWindow() instead of onboardingStore.openOnboarding()
+    // because this can be called from a game window, not the main window
+    if (isWalletUnConnected) {
+      showWalletWarning()
+      window.api.focusMainWindow()
+      window.api.openOnboarding()
+      return
+    }
```
BrettCleary, line 82: I don't think this if statement will ever execute. I tried disconnecting the wallet and signing out and in. The external sync button did not show unless I was signed in and had my wallet connected

src/frontend/store/AlertStore.ts
BrettCleary, line 1: can you write a test on how to get the alert to show?

BrettCleary (commented):
nice work! found a couple small things to address

src/frontend/hooks/useSyncPlayStreakWithExternalSource.ts
eliobricenov: you're right! we don't need this anymore

src/frontend/store/AlertStore.ts
eliobricenov: the alert won't show up anymore, I will remove that

BrettCleary (approved):
lgtm. nice!

src/frontend/ExtensionHandler/index.tsx
flavioislima: Can't we extend the original type on a new type modifying the needed props?

nyghtstalker (approved):
Before signing my playstreak was 0/7 and after signing my playstreak was set to 7/7. I was then able to claim the "HP ERC20 Testing Quest". After claiming the streak returned to 0/7 and was no longer claimable. 

I was able to set the playstreak to 7/7 with the "HP ERC721 Testing Quest", however, I was presented with an error when attempting to claim.

image
image
image
image
image
image

nyghtstalker:
Confirmed that I am now able to claim the ERC721 quests. Great job!

image

src/frontend/ExtensionHandler/index.tsx
eliobricenov: it's a bit complicated and not worth it IMO, wagmi is expecting some TS const assertions, our definitions are similar but without the assertions.

eliobricenov: Actually, I was able to remove the expect error :D

PR #1108, opened 2024-10-09
[UX] Improvements to Update modal on the Store

flavioislima:
Since we got some feedback from users and devs that are not aware of the auto update and thinks that they always need to download the update manually.

It adds 2 new IPC calls that will be used from the store to check the current update status that can be `updating`, `updated` or `idle` (fallback).
Depending on the state we show a different message and action on the store.
If updating, it is possible to close the modal.
If has updated, a button to restart the client is shown.
If idle, we go back to the current behavior of showing the URL to download.

Also, re-enabled the `autoInstallOnAppQuit` option. I think was `false` by mistake.

Store changes: https://github.com/HyperPlay-Gaming/hyperplay-store/pull/223

### How to test

- Download and install the Windows build here from this PR.
- Open HyperPlay and the Update Modal on the Store show should the `Updating` message.
- On the HyperPlay logs screen you can see the download progress messages to confirm the app is being downloaded.
- When asked if you want to restart to conclude the Update, select NO.
- On the Store now the Update Modal should tell the update was completed and show a button to restart the client and apply the update.
- If the client is not updating in the background (test on mac and linux that doesn't have auto update), it should show the current modal we have now.
- Limitations: Messages are not updated automatically on the Store, only when navigating back and forth there.

### Screenshots
image
image

---


package.json
```diff
@@ -54,7 +54,8 @@
       ],
       "files": [
         "build/bin/win32/*"
-      ]
+      ],
+      "verifyUpdateCodeSignature": false
```
flavioislima, line 58: Same.

src/backend/storeManagers/hyperplay/utils.ts
```diff
@@ -219,15 +219,15 @@ export function refreshGameInfoFromHpRelease(
             : data.project_name
         }
       ],
-      storeUrl: `https://store.hyperplay.xyz/game/${data.project_name}`
+      storeUrl: `https://hyperplay-store-git-feat-checkuptapi-hyperplay.vercel.app/game/${data.project_name}`
```
flavioislima, line 222: Will revert before merging.

src/frontend/screens/WebView/index.tsx
```diff
@@ -34,7 +34,10 @@ import cn from 'classnames'
 
 function urlIsHpUrl(url: string) {
   const urlToTest = new URL(url)
-  return urlToTest.hostname === 'store.hyperplay.xyz'
+  return (
+    urlToTest.hostname ===
+    'hyperplay-store-git-feat-checkuptapi-hyperplay.vercel.app'
+  )
```
eliobricenov, line 40: I think we can do this to avoid hardcoding the URL twice
```ts
function urlIsHpUrl(url: string) {
  const urlToTest = new URL(url)
  return (
    urlToTest.hostname ===
    new URL(HYPERPLAY_STORE_URL).hostname
  )
}
```

flavioislima: true

jiyuu-jin (approved):
Looks good!

nyghtstalker (approved):
Confirmed that I was able to get the auto update to work successfully. 

I want to make sure and document my findings so it can be improved on later.

- I couldn't get the auto update to trigger until I unchecked and then rechecked "Check for HP Updates on Startup" option in Settings. After restarting it started the update process.

- A Windows notification (instead of the HP modal included in this PR) appears BEHIND the windows instead of ON TOP of the Window to inform the user that the update has finished and asks the user to restart. I barely noticed it in the toolbar at the bottom! This is very easy to miss... 

image
image

flavioislima:
Yeah, I wonder if updating electron-updater will fix some of the bugs on the electron side. I will continue to improve this flow on future PRs.

flavioislima:

flavioislima:
recheck

flavioislima:

BrettCleary (approved):

PR #1110, opened 2024-10-11
[Fix] Sync playsession for non-hp runners

BrettCleary:
# Summary

Pulls in https://github.com/HyperPlay-Gaming/quests-ui/pull/18

Previously the `Overlay` was conditionally rendered on overlay toggle, causing re-renders that reset the sync timer. Now the overlay is hidden but not removed from the DOM so that the sync hooks won't reset on overlay toggle.

# Testing 
https://app.qase.io/project/HC?suite=75&previewMode=modal&case=368

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hyperplay",
-  "version": "0.20.1",
+  "version": "0.20.2",
   "private": true,
   "main": "build/main/main.js",
   "homepage": "./",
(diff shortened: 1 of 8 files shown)


flavioislima (approved):
Lgtm

eliobricenov (approved):

nyghtstalker (approved):
The overlay is launching for Fortnite and I am seeing the correct logs. I am able to claim the "Test Fortnite Quest" quest.

image
image
image
image

PR #1111, opened 2024-10-16
[Fix] Keep playstreak quests in sync

BrettCleary:
# Summary

Brings in https://github.com/HyperPlay-Gaming/quests-ui/pull/19

desktop client implementation of https://github.com/HyperPlay-Gaming/product-management/issues/657

# Testing

https://app.qase.io/case/HC-378


flavioislima:
recheck

src/frontend/hooks/useKeepPlaystreaksInSync.ts
```diff
@@ -0,0 +1,19 @@
...
+    return
+  }, [syncInitialized])
+  useEffect(() => {
+    async function initSync() {
+      await questPlayStreakSyncState.keepProjectQuestsInSync(appName)
+      setSyncInitialized(true)
+    }
+    initSync()
+  }, [])
+}
```
eliobricenov, line 19: we don't really need a state for this.

```tsx
import { questPlayStreakSyncState } from '@hyperplay/quests-ui'
import { useEffect, useRef } from 'react'

export function useKeepPlaystreaksInSync(appName: string) {
  const syncInitializedRef = useRef(false)

  useEffect(() => {
    if (syncInitializedRef.current) {
      return () => questPlayStreakSyncState.clearAllTimers()
    }
  }, [])

  useEffect(() => {
    const initSync = async () => {
      try {
        await questPlayStreakSyncState.keepProjectQuestsInSync(appName)
        syncInitializedRef.current = true
      } catch (error) {
        console.error('Failed to initialize sync:', error)
      }
    }
    initSync()
  }, [appName])
}
```

eliobricenov, line 19: we don't really need a state for this.

```tsx
import { questPlayStreakSyncState } from '@hyperplay/quests-ui'
import { useEffect, useRef } from 'react'

export function useKeepPlaystreaksInSync(appName: string) {
  const syncInitializedRef = useRef(false)

  useEffect(() => {
    if (syncInitializedRef.current) {
      return () => questPlayStreakSyncState.clearAllTimers()
    }
  }, [])

  useEffect(() => {
    const initSync = async () => {
      try {
        await questPlayStreakSyncState.keepProjectQuestsInSync(appName)
        syncInitializedRef.current = true
      } catch (error) {
        console.error('Failed to initialize sync:', error)
      }
    }
    initSync()
  }, [appName])
}
```

src/frontend/hooks/useKeepPlaystreaksInSync.ts
```diff
@@ -0,0 +1,36 @@
...
+      if (syncInitializedRef.current) {
+        questPlayStreakSyncState.clearAllTimers()
+      }
+    }
+  }, [])
+
+  useEffect(() => {
+    const initSync = async () => {
+      try {
+        questPlayStreakSyncState.init({
```
BrettCleary, line 20: This is the only place it is used, so it is simpler to just call init right before. I initially explored adding it to `StoreController`, but the `StoreController` `useEffect` executes after this useEffect, which breaks `keepProjectQuestsInSync`

src/frontend/hooks/useKeepPlaystreaksInSync.ts
```diff
@@ -0,0 +1,36 @@
...
+          getQuests: window.api.getQuests,
+          getQuest: window.api.getQuest,
+          getUserPlayStreak: window.api.getUserPlayStreak,
+          syncPlaySession: window.api.syncPlaySession,
+          appQueryClient: queryClient
+        })
+
+        await questPlayStreakSyncState.keepProjectQuestsInSync(appName)
```
BrettCleary, line 28: `clearAllTimers` does not need to be called since the store and timers are destroyed when the overlay is closed, which is also when we should stop syncing play sessions

src/frontend/hooks/useKeepPlaystreaksInSync.ts
eliobricenov: how interesting, in my implementation of this in the store I used the StoreController approach 

https://github.com/HyperPlay-Gaming/hyperplay-store/pull/228/files#diff-c4fdb34f33780698be8e19adc9cda4a65464bcd97bdc4a7ed870e888a3c5e42d

tbh I think this way is simpler and does the work too

eliobricenov (approved):

eliobricenov (approved):

flavioislima (approved):

jiyuu-jin (approved):

src/frontend/hooks/useKeepPlaystreaksInSync.ts
BrettCleary: maybe we should move this into `@hyperplay/quests-ui` if it will be used on the store as well

BrettCleary:
@nyghtstalker Ready for a retest

nyghtstalker (approved):
- Confirmed that quests are moving to the claimable section when ready to be claimed
- Confirmed the claim button activates and is useable when quest is claimable
- Confirmed claim button works and reward can be claimed
- Confirmed that when the timer for the days completion is finished it will add and display the day to the count and be claimable if reaching the set day amount needed.

PR #1120, opened 2024-10-18
[Tech] add utm_source to MetaMask urls

BrettCleary:
# Summary

closes https://github.com/HyperPlay-Gaming/product-management/issues/666

# Testing

Check that portfolio.metamask.io pages are still opening


eliobricenov (approved):

src/frontend/screens/Onboarding/walletSelection/screens/scan/index.tsx
```diff
@@ -40,7 +40,9 @@ const ScanScreen = (props: ScanScreenProps) => {
         )}{' '}
         <a
           onClick={() =>
-            window.api.openExternalUrl('https://metamask.io/download/')
+            window.api.openExternalUrl(
+              'https://metamask.io/download/?utm_source=hyperplay'
```
flavioislima, line 44: Would be good to use this in a exported variable like above to easy future changes if needed.

src/frontend/screens/MetaMaskPortfolio/index.tsx
```diff
@@ -13,7 +13,7 @@ const MetaMaskPortfolio = function () {
         <div className={MetaMaskPortfolioStyles.loader}></div>
         <webview
           className={MetaMaskPortfolioStyles.homeWebview}
-          src={`https://portfolio.metamask.io/${page}`}
+          src={`https://portfolio.metamask.io/${page}?utm_source=hyperplay`}
```
flavioislima, line 16: Maybe would be good to use variables and a URL constructor to handle this on a variable instead of directly here.

BrettCleary: sgtm. I moved it to the constants module as a function. I'd prefer to just keep it simple for now but if the url gets more complicated, we could refactor to URL object

src/frontend/screens/Onboarding/walletSelection/screens/scan/index.tsx
BrettCleary: sgtm. made the change 👍

nyghtstalker (approved):
Confirmed that I am still able to access the portfolio page from MM in various locations.

image
image
image

PR #1132, opened 2024-10-28
feat(quest-details): add claim in quest page

eliobricenov:
Steps for testing:

- On-Chain: https://app.qase.io/case/HC-377
- Points: https://app.qase.io/case/HC-380

closes https://github.com/HyperPlay-Gaming/product-management/issues/623

---


src/frontend/components/UI/QuestDetails/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+  }, [address, questId, isSignedIn])
+
+  const { syncPlayStreakWithExternalSource } =
+    useSyncPlayStreakWithExternalSource({
+      refreshPlayStreak: invalidateQuery
+    })
+
+  return (
+    <QuestDetailsWrapper
+      // @ts-expect-error - TODO: investigate why this component is not using the types from QuestWrapperContextValue
```
eliobricenov, line 95: This is the same thing that's been happening with the ui lib, for some reason the bundler is doing something that prevents the consumer code from detecting the props of extended interfaces/types.
```ts
import { QuestWrapperContextValue } from '../../../../../../../../src/types/quests';
export interface QuestDetailsWrapperProps extends QuestWrapperContextValue {
    selectedQuestId: number | null;
    className?: string;
}
export declare function QuestDetailsWrapper(props: QuestDetailsWrapperProps): import("react/jsx-runtime").JSX.Element;
```

eliobricenov (commented):
I will address this on Monday.

src/frontend/components/UI/QuestDetails/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+import useGetUserPlayStreak from 'frontend/hooks/useGetUserPlayStreak'
+import authState from 'frontend/state/authState'
+import { useFlags } from 'launchdarkly-react-client-sdk'
+import { useTranslation } from 'react-i18next'
+import { useAccount } from 'wagmi'
+import { useSyncPlayStreakWithExternalSource } from 'frontend/hooks/useSyncPlayStreakWithExternalSource'
+
+/**
+ * Don't delete this comment block since it's used for translation parsing for keys that are on the quests-ui library.
+ * As a heads up, everytime you add a new key on any library, you need to add it as a block comment anywhere in the code as well.
```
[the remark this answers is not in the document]
eliobricenov: That's a great idea, actually. I tried adding that pattern to the config, but i18n-next-parser is not detecting the usage of the `useTranslation` because the content of the quests-ui is minified. I guess it is because the package expects a text match for `t('**')`. 

image


From what I see in the docs, the text match is not configurable, and I don't think we could use it anyways because the minified code is not predictable. Looking at the code, the `t('code')` is replaced with different letters in each usage.

image
image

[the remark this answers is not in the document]
eliobricenov: Further investigation is out of the scope for this PR but def worth checking.

src/frontend/screens/Quests/index.tsx
```diff
@@ -35,6 +41,58 @@ export function QuestsPage() {
...
+          return navigate(`/store-page?store-url=${storeRedirectUrl}`)
+        })
+    },
+    onError: (error, variable) => {
+      window.api.logError(
+        `Error navigating to ${variable} game: ${error.message}`
+      )
+    }
+  })
+
```
eliobricenov, line 95: This is the same code from QuestDetailsViewPlay.

package.json
```diff
@@ -169,8 +169,8 @@
     "@fortawesome/react-fontawesome": "^0.2.2",
     "@hyperplay/chains": "^0.3.0",
     "@hyperplay/check-disk-space": "^3.5.2",
-    "@hyperplay/quests-ui": "^0.0.38",
-    "@hyperplay/ui": "^1.8.9",
+    "@hyperplay/quests-ui": "0.1.0-alpha.2",
```
jiyuu-jin, line 172: Just making a note here, we probably want to bump this to a non-alpha version.

jiyuu-jin (approved):
Code looks great, the components are very well structured well broken up 🚀

src/frontend/components/UI/QuestRewardClaimedToast/index.tsx
BrettCleary, line 1: can we add a timeout to disable the toast after 5 seconds? either here or in quests-ui

src/frontend/OverlayManager/Overlay/index.tsx
```diff
@@ -147,6 +149,7 @@ export const Overlay = observer(function ({
           <div className={BrowserGameStyles.rightSideContainer}>
             {exitGameButton}
             {extensionManager}
+            <QuestRewardClaimedToast />
```
BrettCleary, line 152: this won't show if the overlay is hidden. I think this might overlap the existing transaction notification and initial toasts. Let's remove it and explore adding more variety to the overlay toasts on a separate PR

BrettCleary (approved):
LGTM. left a couple small change requests that we should make before moving into testing

src/frontend/components/UI/QuestRewardClaimedToast/index.tsx
eliobricenov: done

src/frontend/OverlayManager/Overlay/index.tsx
eliobricenov: You're right; I think it is fine for now, though. That's the position the design team placed it too: https://www.figma.com/design/GKPELZxk551oVC3EDLgEDA/Overlay?node-id=14764-21449&node-type=frame&m=dev

Screenshot 2024-11-12 at 3 53 10 PM

eliobricenov: We decided to remove it and explore in follow up PRs

src/frontend/components/UI/QuestDetails/index.tsx
eliobricenov: turns out there is an active issue for this https://github.com/qmhc/vite-plugin-dts/issues/330

I time-boxed this to 1hr but I couldn't solve it in time. I created this ticket https://github.com/HyperPlay-Gaming/hyperplay-ui/issues/415, we can discuss this in the next sprint planning.

nyghtstalker (approved):
- Confirmed that an on-chain quest is claimable on the quests page
- Confirmed that a non-chain quest is claimable on the quests page
- Confirmed that success toast appears after claimed (on quests page only)
- Confirmed new message when attempting to claim a reward before it can be claimed
- Confirmed that an on-chain quest is claimable in the overlay
- Confirmed that a non-chain quest is claimable in the overlay
- Confirmed that success toast does NOT appear after claiming in the overlay

image
image
image
image
image
image

PR #1138, opened 2024-10-30
[Fix] ipdt build

BrettCleary:
# Summary

valist sdk was a dev dep on client but a dep on hp patcher so it wasn't being rolled up properly


jiyuu-jin (approved):
LGTM

PR #1142, opened 2024-11-01
[Fix] Improve error logs for license key validate and get auth

BrettCleary:
# Summary 

Improve error logs for license key validate and get auth


eliobricenov (approved):

flavioislima (approved):
Good stuff!

I wonder if we can add something like this to deal with Adding games to library, etc.
We could try to ping the Listings api to check if it will return an error and show that to the user in an error modal.

PR #1149, opened 2024-11-09
[Fix] ext lock connect

BrettCleary:
# Summary

closes https://github.com/HyperPlay-Gaming/product-management/issues/728

# Testing

https://app.qase.io/case/HC-379


src/backend/proxy/providerPreload.ts
```diff
@@ -103,35 +107,46 @@ contextBridge?.exposeInMainWorld('providerApi', providerApi)
...
-        accounts: undefined
-      }
-
-      windowAny.ethereum.on('accountsChanged', (accounts: string[]) => {
-        console.log('accounts changed', accounts)
-        windowAny.ethereum.selectedAddress = accounts[0]
-        windowAny.ethereum.accounts = accounts
-      })
+    const windowAny = window as any
+    windowAny.ethereum = {
```
flavioislima, line 111: Since we have this object here and know the keys and values, maybe we could generate a interface from it and remove the `any` type.

flavioislima (approved):
LGTM!
Left a small comment.

jiyuu-jin:
The wallet popup worked well for me, but I got stuck on `Qase step 4` when I click connect on the Portfolio is still just errors out for me.

image

image

jiyuu-jin:
Werid it only failed the first times I ran through the qase test, but I'm not able to reproduce to screenshot the error I got. All subsequent times I've run through the qase flow from beginning to end it's worked now 🙏 

image

jiyuu-jin (approved):

src/backend/proxy/providerPreload.ts
BrettCleary: Sounds good. `window.ethereum` can be kind of hard to type but I took some time to improve our approach here. We now type it in one spot across the app using `MetaMaskInpageProvider` from `@metamask/providers`

I didn't have time to fix every type mismatch it threw (e.g. saying `selectedAddress` is read only when it is not) but this is a step in the right direction

nyghtstalker (approved):
- Confirmed launching a browser game that has quests (DeFi) with the extension attached but locked - the overlay does open and the portfolio tab is accessible. It can't be used since the wallet is locked and the user is presented with an error after clicking the "Connect Wallet" buttons.
- Confirmed the same thing for a browser game that does not have quests. All the same behavior occurs.
- Confirmed that after the wallet is unlocked and the "Connect Wallet" button is selected the user is able to use the portfolio options in the overlay
- Confirmed the same behavior for the two games above
- Confirmed the same behavior for a native game with wallet locked and unlocked
- Confirmed that after launching Xociety, opening OpenSea in the overlay, unlock MM ext, and then connecting the wallet the user is no longer taken to Chromium page and instead the transaction is sent to MM.


image
image
image
image

PR #1152, opened 2024-11-15
[Fix] UI - overflow of quests container when logged out in overlay

BrettCleary:
# Summary

When logged out of HP, the quests viewer div was overflowing the page and the whole overlay was scrollable.

This keeps the container within the viewport and keeps the scrolling only inside the content of the quests viewer.


jiyuu-jin (approved):

eliobricenov (approved):

nyghtstalker (approved):
Confirmed the page is displaying correctly and the scrollbar is no longer dancing on its own.

image

PR #1156, opened 2024-11-16
[Fix] Native sideload failing because of relative path handler

flavioislima:
Fixed an issue Jacob found when sideloading anything on linux that uses a relative path to the root: for instance: 
`usr/bin/steam` .
Hyperplay fails to launch it because it doesn't get the correct exe path.

### How to test

- This PR is simple but touches the launch command directly, so everything related with lanching should be tested.
- Should be tested on all three platforms
- Test sideloaded games and hyperplay games to see if they will launch fine.
- On Linux test passing things like `usr/bin/steam` or any other software installed on system level as well. 

---


src/backend/storeManagers/storeManagerCommon/games.ts
```diff
@@ -342,6 +342,11 @@ export async function launchGame(
 
     // Native
     if (isNative) {
+      const isAbsolutePath = path.isAbsolute(executable)
+      const exeOnly = isAbsolutePath
```
BrettCleary, line 346: looks like this is shadowing the outer scope `exeOnly` variable from 
`const { launchArgs, executable: exeOnly } = getExecutableAndArgs(executable)`

Could we rename this variable for readability?

Also could we just reuse the `exeOnly` variable in the outer scope instead of calling 
`getExecutableAndArgs(executable).executable` again?

flavioislima: Yeah, it was intentional but I agree its confusing. Will find another way.

flavioislima: Moved the fix to the method `getExecutableAndArgs` in `utils` where it should be from the start.

src/backend/utils.ts
```diff
@@ -1255,6 +1256,10 @@ export function getExecutableAndArgs(executableWithArgs: string): {
   const executable = match ? match[0] : ''
   const launchArgs = executableWithArgs.replace(executable, '').trim()
 
+  if (isAbsolutePath) {
+    return { executable: executableWithArgs, launchArgs }
+  }
+
```
BrettCleary, line 1262: having trouble understanding this one. so if the exe path is an absolute path, then we don't want to return the exe with the launch args stripped?

What happens when a exe with args i.e `"/usr/123/asdf/aaa--arg=1"` is passed to `isAbsolute`? Does it return false or does it throw? Seems that might be an unhandled edge case.

BrettCleary (commented):
Could you explain a bit more where the issue is throwing or failing to launch the game when an absolute path is supplied?

I am also confused because we supply absolute paths to `getExecutableAndArgs` from `isGameAvailable` without issue and then call `existsSync` with that.

It seems to me the root cause lies somewhere else in the codebase

src/backend/utils.ts
flavioislima: If you add some console.logs you will be able to see that the `exeOnly` is coming empty and then it first fails to check permissions and then fails to chmod it and finally fails to launch it.

BrettCleary: sorry I meant to write `"/usr/123/asdf/aaa --arg=1"` with the space. Won't that come back as `isAbsolutePath` false and then return `{ executable: "/usr/123/asdf/aaa --arg=1"...` ?

BrettCleary: Also could we add tests for this function since it is a pure utils function?

eliobricenov (approved):
lgtm, let's add another test case for this in the getExecutableAndArgs test suite

src/backend/utils.ts
flavioislima: I see, yes, that was one good edge case. should be handled now.

nyghtstalker (approved):
I did the following for all OSs and no issues were found.
- Launching native games are good
- Launching browser games are good
- Sideloading and launching native games are good
- Sideloading and launching browser games are good

PR #1160, opened 2024-11-20
[UX] Update compatibility dialog to be non-modal

BrettCleary:
# Summary

closes https://github.com/HyperPlay-Gaming/product-management/issues/732

updates the `Dialog` component to accept a `isModal` property. Set it to true for compatibility layer warning so MM ext can still be interacted with at this stage.

`isModal` defaults to true to keep the same previous behavior for all other modals.

# Testing

https://app.qase.io/case/HC-381


eliobricenov (approved):
LGTM. Didn't know this was part of the HTML standard 🤯

flavioislima (approved):
I think this looks fine. 
But since we now have a Modal component on the HP-UI perhaps we should replace the current one with that and use it from now on instead of the client implementation since they are similar.

BrettCleary:
> I think this looks fine. But since we now have a Modal component on the HP-UI perhaps we should replace the current one with that and use it from now on instead of the client implementation since they are similar.

I'd rather not bring in that refactor on this branch. We'll also have to refactor `src\components\Modal\ModalAnimation\index.tsx` in hp/ui to be polymorphic so that we can make the root element a `dialog` element as well. It's a pretty big scope increase.

nyghtstalker (approved):
- Confirmed the compatibility layer warning and the MM login modals both appear when the user launches a non-native game.
- Confirmed the user can login to MM and the comp. layer modal is unaffected

image

PR #1162, opened 2024-11-20
[FIX] Update Game status after finishing playing and not logged in

flavioislima:
Fix an issue where the game status won't change after finishing playing and the player is not logged in on HyperPlay.
The issue was because it was getting stuck on the `syncPlaySession` method and was never triggering the `gameStatusUpdate` with the `done` status.

Moved this one to be above the playtime and also added some `try/catch` to the sync methods.

---

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -1332,6 +1332,12 @@ ipcMain.handle(
     await addRecentGame(game)
 
     if (autoSyncSaves && isOnline()) {
+      /**
+       * @dev It sets to done, so the GlobalState knows that the game session stopped.
+       * Then it changes the status to syncing-saves. Then It sets to done again.
+       * Otherwise it would count the Syncing Saves time (which can be long depending on the game) as playing time as well.
+       * done is not only the state for stopping playing but for finishing any other process that came before.
+       */
       sendFrontendMessage('gameStatusUpdate', {
         appName,
         runner,
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -1111,7 +1111,21 @@ function startNewPlaySession(appName: string) {
   return prevStartTime
 }
 
-async function syncPlaySession(appName: string, runner: Runner) {
+async function postPlaySession(
+  appName: string,
+  runner: Runner,
+  sessionPlaytimeInMs: bigint
+) {
+  const game = gameManagerMap[runner].getGameInfo(appName)
+  const { hyperPlayListing } = await gameIsEpicForwarderOnHyperPlay(game)
+  await postPlaySessionTime(
+    hyperPlayListing?.project_id || appName,
+    // round up to prevent session time loss
+    parseInt(((sessionPlaytimeInMs + BigInt(1000)) / BigInt(1000)).toString())
+  )
+}
+
+function syncPlaySession(appName: string) {
   if (!Object.hasOwn(gamePlaySessionStartTimes, appName)) {
     return BigInt(0)
   }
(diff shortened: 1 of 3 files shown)


eliobricenov (approved):

jiyuu-jin (approved):

src/backend/main.ts
```diff
@@ -1362,6 +1366,17 @@ ipcMain.handle(
       status: 'done'
     })
 
+    // Playtime of this session in minutes. Uses hrtime for monotonic timer not subject to clock drift or sync errors
+    let sessionPlaytimeInMs: bigint | undefined = undefined
+    try {
+      sessionPlaytimeInMs = await syncPlaySession(appName, runner)
```
BrettCleary, line 1372: We should still get the session playtime if the post fails so that the game closed track event still works

BrettCleary: If the player is not logged in, how does it fail? Does it throw or is it stuck waiting?

src/backend/main.ts
```diff
@@ -1362,6 +1366,17 @@ ipcMain.handle(
       status: 'done'
     })
 
+    // Playtime of this session in minutes. Uses hrtime for monotonic timer not subject to clock drift or sync errors
```
BrettCleary, line 1369: Should change this from minutes to ms

src/backend/main.ts
```diff
@@ -1339,6 +1337,12 @@ ipcMain.handle(
         status: 'syncing-saves'
       })
 
+      sendFrontendMessage('gameStatusUpdate', {
```
BrettCleary, line 1340: This is duplicated from above

BrettCleary (commented):
Can we bump the package.json version on this branch too?

src/backend/main.ts
flavioislima: done

src/backend/main.ts
flavioislima: I dont think it will throw anything tbh becase I dont see how it might fail. I think I can revert this to the original code.

src/backend/main.ts
flavioislima: It is not actually.
It sets to `done`, so the GlobalState knows that the game session stopped.
Then it changes the status to `syncing-saves`. Then It sets to `done` again.
So, `done` is not only the state for stopping playing but for finishing any other process that came before.
So this is correct.

BrettCleary: Could you add a comment explaining this here?

src/backend/main.ts
BrettCleary: I was thinking something more like this https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/pull/1165 so that we still send the game closed event even if they aren't logged in

BrettCleary:
opened back up this comment btw https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/pull/1162#discussion_r1855877976

src/backend/main.ts
BrettCleary: added the comment

BrettCleary (approved):
LGTM

nyghtstalker (approved):
-- While the user is LOGGED OUT
- Confirmed that the Play button is accessible after closing a native game
- Confirmed that the Play button is accessible after closing a browser game

-- While the user is LOGGED IN
- Confirmed that the Play button is accessible after closing a native game
- Confirmed that the Play button is accessible after closing a browser game

PR #1168, opened 2024-11-26
feat(quest-page): add auto expand and tab filters

eliobricenov:
### Summary

closes https://github.com/HyperPlay-Gaming/product-management/issues/627

### How to test:

- go to the quest page
image
- check quest detail is auto expanded 
image

---


src/frontend/screens/Quests/components/QuestsSummaryTableWrapper/index.tsx
eliobricenov, line 1: The changes are basically moving the filter/sort state from this component to one level above. Once the filtering is the parent component, we can default to the first quest **after** filtering/sorting and pass the selected quest id to the quest list and quest detail components.

eliobricenov: After moving the state, the wrapper component is pretty much empty so I decided to remove it and implement QuestSummaryTable directly in the quest page component.

jiyuu-jin (approved):
Worked and code LGTM

image

BrettCleary (approved):
LGTM. There is a short period on the initial load where the quest details wrapper is not shown, then some layout shift once the quests are fetched. It would be nice if the quest details component was in a loading state at this point too to prevent layout shift.

Here is what it looks like
image


There also seems to be a regression in the quest details background color and border either on this PR or a prior PR

On this PR:
image

On figma: https://www.figma.com/design/yHrBPoE70k1I2icW4paEit/Quests?node-id=1-17948&t=tBaRuRLtOpcveXar-4

nyghtstalker (approved):
Quests are now always auto-expanded by default. 

image

PR #1169, opened 2024-11-26
[TECH] Add more logs and error handling for MW Mod support

flavioislima:
- Adds a timeout while waiting for the files to be copied to the upper directory, this is necessary since some people are getting stuck forever on this step.
- Adds some permission checks to the source and destination files.
- Adds some try/catches to throw errors while extracting, copying, moving files, etc.
- Adds some tests to the `copyRecursiveAsync` method.
- Log more verbosely.
- Needed to update jest config and dependencies to handle the creating and deleting of files and folders and to deal with the timeout for the method above.

### HOW TO TEST

- Windows only
- Install Market Wars as usual to see if it will fail or suceed.
- These changes won't change the logic of the patching, only add error handlers.

---


jiyuu-jin (approved):
LGTM

src/backend/utils.ts
```diff
@@ -1306,19 +1306,76 @@ export const writeConfig = (appName: string, config: Partial<AppSettings>) => {
...
+    const destDir = dirname(dest)
+    try {
+      await access(destDir, constants.W_OK)
+    } catch {
+      throw new Error(
+        `No write permission for destination directory: ${destDir}`
+      )
+    }
+  } catch (error) {
+    if (error instanceof Error) {
```
eliobricenov, line 1325: I don't think we need this check, error will always be an instances of the Error class by default

eliobricenov (approved):
lgtm, I only have one comment

src/backend/utils.ts
flavioislima: This is necessary to get the error type properly, without it we dont have access to `error.message`.
But I changed it since in this case would be good to get the whole Error object.

src/backend/__tests__/utils.test.ts
```diff
@@ -1,5 +1,11 @@
 import * as utils from '../utils'
-import { getExecutableAndArgs } from '../utils'
+import { getExecutableAndArgs, copyRecursiveAsync } from '../utils'
+import { mkdir, writeFile, symlink } from 'fs/promises'
+import { join } from 'path'
+import { chmod, existsSync } from 'fs'
```
BrettCleary, line 5: lint error here

src/backend/utils.ts
```diff
@@ -1306,9 +1306,16 @@ export const writeConfig = (appName: string, config: Partial<AppSettings>) => {
   }
 }
 
+const COPY_TIMEOUT_MS = 30000 // wait time before throwing a timeout error
```
BrettCleary, line 1309: we won't exceed this during normal operation on slow machines will we?

src/backend/__tests__/utils.test.ts
```diff
@@ -261,4 +267,77 @@ describe('backend/utils.ts', () => {
...
+      const linkFile = join(sourceDir, 'link.txt')
+      await writeFile(testFile, 'test content')
+      await symlink(testFile, linkFile)
+
+      await copyRecursiveAsync(linkFile, join(destDir, 'link.txt'))
+
+      expect(existsSync(join(destDir, 'link.txt'))).toBe(false)
+    })
+
+    it('should throw on timeout', async () => {
```
BrettCleary, line 319: Could we use a fake timer here so that the jest tests run time is not increased by 30 seconds? https://jestjs.io/docs/timer-mocks#advance-timers-by-time

src/backend/utils.ts
flavioislima: I dont think so, that is actually a pretty generous time limit just to avoid being stuck forever. the file would need to be huge to exceed this imo.

src/backend/__tests__/utils.test.ts
flavioislima: fixed

nyghtstalker (approved):
Confirmed the game is now installed at the correct location and launching.

PR #1170, opened 2024-11-26
rm global class from weclome modal lang selector

BrettCleary:
# Summary

A global class intended for the lang selector in settings was breaking the lang selector on the welcome modal at screen sizes <850px

Before
image

After
image


biliesilva (approved):
perfect, ty!

PR #1172, opened 2024-11-26
Fix/game closed event on main win close

BrettCleary:
# Summary

Give game launch handlers 1 second on main window close to clean up (i.e. sync playsession and send game closed event).

closes https://github.com/HyperPlay-Gaming/product-management/issues/714

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -4,6 +4,7 @@ import {
   AppSettings,
   GamepadInputEvent,
   GameSettings,
+  LaunchParams,
   Runner,
   StatusPromise
 } from 'common/types'
@@ -173,7 +174,7 @@ import 'backend/proxy/ipcHandlers'
 import './ipcHandlers'
 import './ipcHandlers/checkDiskSpace'
 
-import { metricsAreEnabled, trackEvent } from './metrics/metrics'
+import { flushEvents, metricsAreEnabled, trackEvent } from './metrics/metrics'
 import { hpLibraryStore } from './storeManagers/hyperplay/electronStore'
 import { libraryStore as sideloadLibraryStore } from 'backend/storeManagers/sideload/electronStores'
 import { backendEvents } from 'backend/backend_events'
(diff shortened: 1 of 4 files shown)


jiyuu-jin (approved):
code looks good.

src/backend/main.ts
```diff
@@ -1161,228 +1182,231 @@ ipcMain.handle('isClientUpdating', async () => {
 
 ipcMain.on('restartClient', () => autoUpdater.quitAndInstall())
 
-// get pid/tid on launch and inject
-ipcMain.handle(
```
BrettCleary, line 1165: i basically moved all of this logic out into `handleGameLaunch` so that we could store the promise in `allLaunchHandlerPromises` to await on app quit

src/backend/metrics/metrics.ts
```diff
@@ -161,6 +161,8 @@ export const trackEvent = async ({
   }
 }
 
+export const flushEvents = async () => rudderstack.flush()
```
BrettCleary, line 164: we have to flush the events in the launch handler so that `Game Closed` is sent on app close. Otherwise it is added to the cache and will wait a little while to see if there are other events that can be batch sent along with it

src/backend/utils.ts
```diff
@@ -256,9 +257,12 @@ async function handleExit() {
         logInfo([`Unable to kill ${procName}, ignoring.`, error])
       }
     })
+  }
+  // Kill all child processes, closing browser and native games that are running
+  callAllAbortControllers()
```
BrettCleary, line 262: this has to be called before the cleanUp function so that that launch handlers can finish executing

src/backend/utils/aborthandler/aborthandler.ts
```diff
@@ -29,7 +29,7 @@ function callAbortController(id: string) {
 }
 
 function callAllAbortControllers() {
-  for (const key in abortControllers.keys()) {
```
BrettCleary, line 32: This was causing native games launched through HP to not be closed when HP was closed

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "hyperplay",
-  "version": "0.22.1",
+  "version": "0.22.2-analytics",
```
BrettCleary, line 3: just for testing so I can filter for this version on the charts

BrettCleary:
Pretty sensitive code. Will need your review on this one @flavioislima

flavioislima (approved):
Code looks good! left a small comment.

Did some basic testing with Epic and GOG games to check if the sync-saves would work on app close and they are working as they should.

Ideally would be good to test the launch command pretty well just to guarantee nothing is broken on any platform for any runner.

PR #1183, opened 2024-12-06
[Feat] Filter out ineligible completed quests

BrettCleary:
# Summary

Support quests with completed status

closes https://github.com/HyperPlay-Gaming/product-management/issues/531

# Testing

1. Not signed in doesn't show a completed status quest
2. Signed in and meeting eligibility requirements shows completed status quest
3. Signed in and not meeting eligibility requirements does not show completed status quest


src/backend/ipcHandlers/quests.ts
```diff
@@ -10,8 +10,14 @@ import {
...
-  let url = `${DEV_PORTAL_URL}api/v1/quests?questStatus=ACTIVE`
+async function fetchQuests({
+  projectId,
+  status
+}: {
+  projectId?: string
+  status: 'ACTIVE' | 'COMPLETED'
+}): Promise<Quest[]> {
+  let url = `${DEV_PORTAL_URL}api/v1/quests?questStatus=${status}`
   if (projectId) {
```
jiyuu-jin, line 21: Just a small nit pick but always better to use new Url() when adding parameters.

src/frontend/components/UI/QuestsViewer/components/QuestLogWrapper/index.tsx
```diff
@@ -9,38 +9,37 @@ import styles from './index.module.scss'
...
-  const quests = questsResults?.data?.data
   const userCredits = useGetG7UserCredits()
   const userCreditsBalance = userCredits?.data?.data
   const pointsBalancesQuery = useGetPointsBalancesForProject(appName)
   const pointsBalances = pointsBalancesQuery?.data?.data
   const { t } = useTranslation()
-  const { questIdToQuestStateMap, isPending: isGetQuestStatesPending } =
-    useGetQuestStates({
-      quests
-    })
```
eliobricenov, line 37: why did you decide to move this hook one level up and remove the quest result from the props? I'm ok with passing quests and isLoading but I'm not sure I understand why we need questIdToQuestStateMap as prop.

We will eventually need to move this component to the quest-ui too as we need these changes in the browser platform too. In that case why don't we keep the pattern of having the hooks inside of the component?

BrettCleary: If the quest is completed but not claimable, it shouldn't be shown in the UI as the user cannot interact with it any more. We need it in the level up so we can filter it out and select the first quest from the quest array, which affects what we show in quest details.

I've updated the logic we will share with the store here https://github.com/HyperPlay-Gaming/quests-ui/pull/40 with updated tests. In the store, I figured we might SSR it. If not, I think we should refactor `useGetQuestStates` to be shared in quests-ui at that time. We still don't have many (any?) live quests in the store right now, so let's not block this PR on that basis.

eliobricenov (approved):

nyghtstalker (commented):
1. Confirmed that I am seeing the completed status quest while logged in
2. Confirmed that I am NOT seeing the completed status quest while logged out
3. Confirmed that after claiming a completed status quest the quest is no longer displayed while logged in.

image
image
image

jiyuu-jin (commented):
Great work!

PR #1185, opened 2024-12-08
[Fix] add wagmi patch to fallback to config.connectors

BrettCleary:
# Summary

Wagmi sometimes fails to add the connector details in the current connection to the config object. This breaks the rewards flow in both the client and the overlay in these cases.

The only values set for the current connection connector when this happens are from https://zustand.docs.pmnd.rs/integrations/persisting-store-data#partialize which only stores the uid, name, type, and id in local storage https://github.com/wevm/wagmi/blob/1e0dd8f3f701c11a081b103fc6e2dcdb5b4063c8/packages/core/src/createConfig.ts#L257. Also the account address is set and the status is `reconnecting`. 

I believe the cause is that we have not called wagmi connect to create this connection with wagmi in the quests-ui package. Account.address is set so `connectAsync` is not called here https://github.com/HyperPlay-Gaming/quests-ui/blob/88dcb79e481aefccaf68eb617295bed8f8784c9a/src/components/RewardWrapper/index.tsx#L217-L224. 

https://github.com/HyperPlay-Gaming/quests-ui/pull/41 handles this. It calls `connectAsync` again if `switchChain` doesn't exist on the wagmi connection.

closes https://github.com/HyperPlay-Gaming/product-management/issues/801

# Testing

This test case but only for the main window quest page claiming https://app.qase.io/case/HC-377

Note that this issue only manifests with the artifact and usually only after an app update, so you may need to install an older version of HP, then install this artifact, and then run test case HC-377 to reproduce.


eliobricenov (approved):

flavioislima (approved):

jiyuu-jin (approved):

nyghtstalker (approved):
Confirmed I was able to claim the erc20 test quest successfully.

image
image
image

PR #1194, opened 2024-12-18
[Fix] Clear cache and storage data on new browser game window

flavioislima:
This should clear the cache of the browser windows when launching games,
It is a workaround for some games like Champz that had issues after games updates.

Another idea would be to clear the cache only when cleaning the hyperplay cache but that might not be intuitive for some users.

### How to Test

Not sure since you would need to have an old version of a game loaded, then on the server the game gets updated and then the game launched through hyperplay to trigger the cache clean and be sure that the game would be launched on the latest version.

---


src/backend/storeManagers/storeManagerCommon/games.ts
```diff
@@ -145,14 +146,16 @@ const openNewBrowserGameWindow = async (
   const hpOverlay = await getHpOverlay()
 
   return new Promise((res) => {
+    const partition = `persist:${appName}`
```
BrettCleary, line 149: nice

BrettCleary (approved):
Tested logging into DFK, closing the game, and re-launching. Confirmed I was still logged in on re-launch

jiyuu-jin (approved):
Makes sense. Code looks good! 🚀

ry-animal (approved):

nyghtstalker (approved):
I did the best I could to test this and everything looks good and works well from what I could tell. Good job!

PR #1200, opened 2024-12-19
[UI]fix search bar sizing on Quest page

biliesilva:
<--- Search bar is disappearing when small screens for the Quest page --->

Bug:
image

I have added some styles to add media queries for some small screens.

---


src/frontend/screens/Quests/index.tsx
```diff
@@ -223,8 +223,8 @@ export function QuestsPage() {
             <SearchBar
               searchText={searchText}
               setSearchText={setSearchText}
-              i18n={{ placeholder: 'Search Quest' }}
-              styles={{ container: { margin: '0px 0px 0px auto' } }}
+              i18n={{ placeholder: 'Search' }}
```
BrettCleary, line 226: Might be good to make this a translation string or reuse an existing translation string

src/frontend/screens/Quests/index.module.scss
```diff
@@ -34,20 +34,34 @@
...
-  @media (max-width: 1300px) {
-    width: 540px;
+  //Steam Deck
+  @media (max-width: 1280px) {
+    width: 600px;
+  }
+
+  @media (max-width: 1024px) {
+    width: 390px;
   }
```
BrettCleary, line 50: You could use the preset media queries that we have in the ui lib `src\styles\utilities\_variables.scss` here. That way the breakpoints are more readable and defined in one place if we want to change them across the app. If you ctrl+f for some of them, you should be able to find some examples in the client or the ui lib of their usage.

BrettCleary (approved):
No more overflow! nice job

flavioislima (approved):
LGTM
+1 to what Brett said.

src/frontend/screens/Quests/index.module.scss
biliesilva: I checked these, but any of them match the sizes I tested for this quest screen. To be able to show the search bar and have a minimal size to show the content in each component piece for 1280 and 1024 screens I had to create these. We can add this to the variables in the ui lib and reuse them here or keep like this as I believe we will not use this specificity in other place, wdyt? @BrettCleary

nyghtstalker (approved):
Looks great! No more cutting off the search bar!

image
image

src/frontend/screens/Quests/index.module.scss
BrettCleary: cool. makes sense. let's just use what you have hardcoded here then. that's fine

PR #1213, opened 2025-01-21
[Fix] MetaMask extension notification window sometimes not showing in overlay

BrettCleary:
# Summary 

closes https://github.com/HyperPlay-Gaming/product-management/issues/764

# Testing 

https://app.qase.io/case/HC-530
- note that the expected behavior for step 7 is failing on main

### DIFF
--- a/src/frontend/screens/MetaMaskHome/index.tsx
+++ b/src/frontend/screens/MetaMaskHome/index.tsx
@@ -24,6 +24,7 @@ const MetaMaskHome = function ({ path = 'home.html' }: MetaMaskHomeProps) {
             ''
           )}`}
           allowpopups={trueAsStr}
+          key={'metamask_home' + path + hash}
         />
       </div>
     </>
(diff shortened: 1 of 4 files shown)


src/frontend/screens/MetaMaskHome/index.tsx
```diff
@@ -24,6 +24,7 @@ const MetaMaskHome = function ({ path = 'home.html' }: MetaMaskHomeProps) {
             ''
           )}`}
           allowpopups={trueAsStr}
+          key={'metamask_home' + path + hash}
```
BrettCleary, line 27: this isn't strictly necessary for this specific issue's fix, but a similar bug could occur so I'm going to go ahead and add the same fix

flavioislima (approved):
LGTM

nyghtstalker (approved):
Confirmed I am getting the extension notifications in the overlay when:
-  Connecting the current wallet
- Adding a new wallet account
- Connecting the newly added wallet account after
- Running the curl command

 image
image

PR #1219, opened 2025-01-23
feat(logger): add support for sending logs to sentry

eliobricenov:
Enhance the logging functionality to support additional options and Sentry integration.

Logging Enhancements:
* [`src/backend/api/misc.ts`](diffhunk://#diff-6a28226c61f9980ba555d23669542c4e555ec104c952ce473312c45179eb7e33L51-R52): Enhanced `logError` function to accept an optional `LogOptions` parameter. [[1]](diffhunk://#diff-6a28226c61f9980ba555d23669542c4e555ec104c952ce473312c45179eb7e33L51-R52) [[2]](diffhunk://#diff-6a28226c61f9980ba555d23669542c4e555ec104c952ce473312c45179eb7e33R107)
* [`src/backend/logger/logger.ts`](diffhunk://#diff-3b36a7de185377bbb618ac5e7bc20b0780c6ea78b5d5ac53c7b9a7469448a988R14-R15): Integrated Sentry for error logging and extended `LogOptions` with additional properties for Sentry configuration. [[1]](diffhunk://#diff-3b36a7de185377bbb618ac5e7bc20b0780c6ea78b5d5ac53c7b9a7469448a988R14-R15) [[2]](diffhunk://#diff-3b36a7de185377bbb618ac5e7bc20b0780c6ea78b5d5ac53c7b9a7469448a988L48-R50) [[3]](diffhunk://#diff-3b36a7de185377bbb618ac5e7bc20b0780c6ea78b5d5ac53c7b9a7469448a988L241-R277)
* [`src/common/types.ts`](diffhunk://#diff-5328ecb0af4d24ae2b95bb12aa89dc1feab5964c8790126d0f721dacc3a99ef9R16-R21): Imported and re-exported `LogPrefix` and `LogOptions` from the logger module.

closes https://github.com/HyperPlay-Gaming/product-management/issues/876

---

- [x] Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)


package.json
```diff
@@ -169,9 +169,9 @@
     "@fortawesome/react-fontawesome": "^0.2.2",
     "@hyperplay/chains": "^0.3.0",
     "@hyperplay/check-disk-space": "^3.5.2",
-    "@hyperplay/quests-ui": "^0.1.9",
+    "@hyperplay/quests-ui": "0.1.12-alpha.4",
     "@hyperplay/ui": "^1.8.18",
-    "@hyperplay/utils": "^0.3.6",
+    "@hyperplay/utils": "0.3.7-alpha.1",
```
eliobricenov, line 174: **reminder: use the final package version once they are live**

BrettCleary (approved):
LGTM

flavioislima (approved):
Good stuff. That was exactly what I had in mind. 🤘

Only thing I would change would be to not use Sentry names explicitly. Because if we want to migrate to another service in the future we would need to rename it.

So a more generic name for the properties would be better IMO. But it's not a blocker.

eliobricenov:
> Good stuff. That was exactly what I had in mind. 🤘
> 
> Only thing I would change would be to not use Sentry names explicitly. Because if we want to migrate to another service in the future we would need to rename it.
> 
> So a more generic name for the properties would be better IMO. But it's not a blocker.

That makes sense; we can revisit it later if we need to do that.

eliobricenov:
Merging as I've confirmed that sentry errors are being spawned https://hyperplay-mm.sentry.io/issues/6238719139/?project=4505435657601024&query=&referrer=issue-stream&sort=freq&statsPeriod=24h&stream_index=0

ry-animal:
late to the show but LGTM lol

PR #1224, opened 2025-01-30
[Fix] Non-MM Extension injected provider in main window and overlay

BrettCleary:
# Summary

closes https://github.com/HyperPlay-Gaming/product-management/issues/863

# Testing
https://app.qase.io/case/HC-377 for wc, mmm, and mm ext

https://app.qase.io/case/HC-529


ry-animal (approved):
Awesome!

src/frontend/state/InjectedProviderHandler.ts
```diff
@@ -0,0 +1,159 @@
...
+
+  async handleRequest(
+    event: Electron.IpcRendererEvent,
+    id: number,
+    args: Record<string, unknown>
+  ) {
+    try {
+      await this.waitForConnection()
+      /* eslint-disable-next-line */
+      //@ts-ignore
```
flavioislima, line 14: Nitty-picky but ts-ignore should have a reason after it, then you won't need the eslint-disable-next-line above it.

src/frontend/state/InjectedProviderHandler.ts
```diff
@@ -0,0 +1,159 @@
...
+      const value = await window.ethereum?.request(args)
+      window.api.returnExtensionRequest(id, value)
+    } catch (err) {
+      console.error(`error during request: ${err}`)
+      window.api.errorExtensionRequest(id, err)
```
flavioislima, line 19: Is this going to sentry as well?

src/frontend/state/InjectedProviderHandler.ts
```diff
@@ -0,0 +1,159 @@
...
+  async waitForConnection(): Promise<void> {
+    const isConnected = window.ethereum?.isConnected()
+
+    if (isConnected) {
+      return
+    }
+
+    return new Promise((resolve) => {
+      /* eslint-disable-next-line */
+      //@ts-ignore
```
flavioislima, line 32: same comment as above.

src/frontend/state/InjectedProviderHandler.ts
```diff
@@ -0,0 +1,159 @@
...
+      removeSendListener()
+      removeSendAsyncListener()
+    }
+  }
+
+  constructor() {
+    makeAutoObservable(this)
+    const bindListeners = (): boolean => {
+      if (typeof window.ethereum !== 'undefined') {
+        console.log('binding ethereum listeners')
```
flavioislima, line 138: Why instead of using all those console.logs, that would be complicated to debug if an user has some issues, you don't call `window.api.logInfo` (logError, etc)? So then it would be included in the app logs that the users would send when opening an issue on zendesk.

BrettCleary: This is in the browser window, so we need to be able to see the logs in that particular browser window. If we were using this in multiple browser windows and only logging to the main node.js process console, it would be impossible to debug. I can do both though so we can see if at least one of the windows binded listeners to window.ethereum

BrettCleary: updated

src/frontend/state/InjectedProviderHandler.ts
BrettCleary: good call

src/frontend/state/InjectedProviderHandler.ts
BrettCleary: No. This returns the error to whoever called this provider whether that be a proxy server call or a window.ethereum call somewhere in the app

flavioislima (approved):
Looks good! 👍

BrettCleary:
ran for all wallet connection methods in the client main window and overlay window 

https://app.qase.io/run/HC/dashboard/8

BrettCleary:
ran another test here https://app.qase.io/run/HC/dashboard/11

PR #1231, opened 2025-02-19
[Fix] Overlay performance issues with MetaMask Extension

BrettCleary:
# Summary

Some games like Kokodi had mouse glitch and other performance issues due to excessive MetaMask popup window processing in the background when the overlay was hidden.

Now we remove the popup/notification windows from the overlay while the overlay is hidden.

The tradeoff is there is a slight ~1 second load time before the user can use the MM extension or notification window after toggling the overlay on.

# Testing

### DIFF
--- a/src/frontend/ExtensionManager/index.tsx
+++ b/src/frontend/ExtensionManager/index.tsx
@@ -42,8 +42,17 @@ const ExtensionManager = function () {
     rootRef.current?.close()
   }
 
+  /**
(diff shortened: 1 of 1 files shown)


BrettCleary:
https://app.qase.io/run/HC/dashboard/13

biliesilva (approved):
Tested on Windows and the game worked ok. I could interact with the overlay and MetMask extension whithout any glitches! :)

flavioislima (approved):

eliobricenov (approved):

PR #1247, opened 2025-03-04
[Feat] implemented Fuse.js to search by quest name and game title.

biliesilva:
## Description

This PR is to add fuzzy search to the client.  It was implemented only in the store and generated bugs because of conflicts of different search logic.

### What to test

 - Test if the search for the **_game name_** matches what is shown and suggested;
 - Test if the search for the **_quest name_** matches what is shown and suggested.

---

### DIFF
--- a/src/frontend/screens/Quests/index.tsx
+++ b/src/frontend/screens/Quests/index.tsx
@@ -144,16 +144,20 @@ export function QuestsPage() {
     }) ?? []
 
   const fuse = new Fuse(questsWithGameNames ?? [], fuseOptions)
-  const searchFilteredQuests = searchText
-    ? (fuse.search(searchText) as FuseResult<Quest>[])
-    : quests?.map((quest) => ({
-        id: quest.id,
-        project_id: quest.project_id,
-        name: quest.name,
-        title: listings?.[quest.project_id]?.project_meta?.name ?? '',
-        item: quest,
-        refIndex: 0
-      })) ?? []
+
+  let searchFilteredQuests: FuseResult<Quest>[] = []
+  if (searchText) {
+    searchFilteredQuests = fuse.search(searchText) as FuseResult<Quest>[]
+  } else if (quests) {
+    searchFilteredQuests = quests.map((quest) => ({
+      id: quest.id,
+      project_id: quest.project_id,
+      name: quest.name,
+      title: listings?.[quest.project_id]?.project_meta?.name ?? '',
+      item: quest,
+      refIndex: 0
+    }))
+  }
 
   const searchQuests = (quests: Quest[], query: string) => {
     if (!query) return quests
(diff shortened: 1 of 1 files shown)


src/frontend/screens/Quests/index.tsx
```diff
@@ -117,63 +127,100 @@ export function QuestsPage() {
...
-        ? listings[project_id]?.project_meta?.main_capsule
+        ? listings[result.item.project_id]?.project_meta?.main_capsule
         : ''
       if (imageUrl) {
         imagesToPreload.push(imageUrl)
       }
-      const title = listings ? listings[project_id]?.project_meta?.name : ''
+      const title = listings
+        ? listings[result.item.project_id]?.project_meta?.name
+        : ''
```
BrettCleary, line 198:
```suggestion
        : ''
        const id = result.item.id
```
that way you don't have to copy paste result.item.id everywhere id was and easier to change in the future

src/frontend/screens/Quests/index.tsx
```diff
@@ -117,63 +127,100 @@ export function QuestsPage() {
...
+  const searchFilteredQuests = searchText
+    ? (fuse.search(searchText) as FuseResult<Quest>[])
+    : quests?.map((quest) => ({
+        id: quest.id,
+        project_id: quest.project_id,
+        name: quest.name,
+        title: listings?.[quest.project_id]?.project_meta?.name ?? '',
+        item: quest,
+        refIndex: 0
+      })) ?? []
```
BrettCleary, line 156:
```suggestion
  let searchFilteredQuests = []
  if (searchText){
    searchFilteredQuests = fuse.search(searchText) as FuseResult<Quest>[]
  } else if (quests) {
    searchFilteredQuests = quests.map((quest) => ({
        id: quest.id,
        project_id: quest.project_id,
        name: quest.name,
        title: listings?.[quest.project_id]?.project_meta?.name ?? '',
        item: quest,
        refIndex: 0
      }))
  }
```

this is more readable

src/frontend/screens/Quests/index.tsx
```diff
@@ -117,63 +127,100 @@ export function QuestsPage() {
...
       )
-    return questTitleMatch || gameTitleMatch || searchByKeywords
-  })
+      if (!currentQuestExists) {
+        const firstQuest = filteredQuests[0]
+        if (firstQuest) {
+          const searchParams = new URLSearchParams(window.location.search)
+          const newUrl = `/quests/${firstQuest.id}${
+            searchParams.toString() ? `?${searchParams.toString()}` : ''
+          }#card-${firstQuest.id}`
```
BrettCleary, line 177: do we need the `#card-${firstQuest.id}` here?

biliesilva: It makes the page scroll to the quest when we search and the page is loaded, this is more used when there is a big list of quests with the same initials for example.

flavioislima (approved):
LGTM

BrettCleary (approved):
LGTM and it works :fire:

nyghtstalker (approved):
Confirmed I can search by the following without issues and will return the correct results:
- Quest title
- Quest description
- Game name

PR #1249, opened 2025-03-05
[FIX] Reload app for MM only on app update

flavioislima:
This pull request includes changes to improve the client update handling and extension initialization process. The most important changes include adding a new function to reload for MetaMask, modifying the frontend to reload the app when a new version is detected, and restructuring the backend to handle the new reload functionality.

Changes to client update handling:

* [`src/backend/api/misc.ts`](diffhunk://#diff-6a28226c61f9980ba555d23669542c4e555ec104c952ce473312c45179eb7e33R241-R242): Added a new `reloadForMM` function to send a reload request for MetaMask.
* [`src/frontend/components/UI/AppVersion/index.tsx`](diffhunk://#diff-82f56da14f774b3ec0e8dcdf571564cd64960f152f133299cbed108c6ce49549L4-R18): Updated the `AppVersion` component to check for version changes and call `reloadForMM` if a new version is detected.

Changes to extension initialization:

* [`src/backend/main.ts`](diffhunk://#diff-859af3c776da914ebff60fa13adfc96e79d99ea563fe1e4e790de14656442accL709-L712): Removed the extension initialization from the `frontendReady` event and added a new `reloadForMM` event to handle the initialization and reloading process. [[1]](diffhunk://#diff-859af3c776da914ebff60fa13adfc96e79d99ea563fe1e4e790de14656442accL709-L712) [[2]](diffhunk://#diff-859af3c776da914ebff60fa13adfc96e79d99ea563fe1e4e790de14656442accR718-R727)

### HOW TO TEST

Issue is that on the current version the app reloads the interface after a few seconds (~5s) every time it's opened. With 
this change it will only do that after an app update.

- Run the App (Dev mode or artifact on any platform)
- The app won't reload
- Open Dev tools  and go to > Applications > Local Storage and change the `last_version` field to a previous version (< 0.23.2)
- Close the app and open it again.
- The app should reload now after a few seconds.

---


biliesilva (approved):

src/frontend/components/UI/AppVersion/index.tsx
```diff
@@ -1,16 +1,21 @@
...
   useEffect(() => {
     window.api.getAppVersion().then((version) => {
+      if (version !== lastVersion) {
+        window.api.logInfo(
+          `Updated to a new version, reloading the MetaMask extension. Old version: ${lastVersion}, new version: ${version}`
+        )
+        window.api.reloadForMM()
+      } else {
+        window.api.logInfo('No new version detected. Wont reload the app')
+      }
```
BrettCleary, line 18: can we do this in the main process instead? these are all backend functions so I think it makes more sense to call them there

BrettCleary (changes requested):
let's move the logic into a function that is only called in the main process

src/frontend/components/UI/AppVersion/index.tsx
flavioislima: It is like this because we are using the localStorage, to be only on the main process will need a bit more refactor.
I would need to remove those methods that are not used anymore and also implement the main process one using electron-store instead. 
Doable, but I didn't want to spend too much time on this one tbh. Especially since this already works fine.

flavioislima: Pushed a commit to start storing the app version on the config store so in a couple of versions we just need to move the content of the `reloadForMM` to that block in `main.js`  and remove the code on the frontend.

BrettCleary: Yeah sounds good. I would just hate to have a long lived coupling/dependency of the functioning of mm extension with the presence of this `AppVersion` UI component in the DOM

BrettCleary:
I believe this would be the correct test case https://app.qase.io/case/HC-529. We should actually go back to 0.22.4 so a mm ext version changes https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/releases/tag/v0.22.4

BrettCleary (approved):
lgtm. appreciate you migrating it to the backend over time. responded on the other comment thread 👍 

Tested upgrading from v0.22.4 and it worked

flavioislima:
Closed due to https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/pull/1261

PR #1251, opened 2025-03-06
[Tech] Use default user agent for hp urls

BrettCleary:
# Summary

Let's use the default user agent for any store.hyperplay.xyz urls

### DIFF
--- a/src/frontend/components/UI/WebviewControls/index.tsx
+++ b/src/frontend/components/UI/WebviewControls/index.tsx
@@ -91,7 +91,7 @@ export default function WebviewControls({
 
   const _url = url !== '' ? new URL(url) : null
   const allowList = [
-    'store.hyperplay.xyz',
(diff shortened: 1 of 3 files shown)


src/frontend/components/UI/WebviewControls/index.tsx
```diff
@@ -91,7 +91,7 @@ export default function WebviewControls({
 
   const _url = url !== '' ? new URL(url) : null
   const allowList = [
-    'store.hyperplay.xyz',
+    'hyperplay-store-git-tech-refactorlaunchercheck-hyperplay.vercel.app',
```
BrettCleary, line 94: note to myself to revert before merging

biliesilva (approved):

BrettCleary:
looks like plausible doesn't support electron as a separate browser unfortunately so let's keep the existing implementation https://github.com/plausible/analytics/discussions/1097

PR #1254, opened 2025-03-10
[Fix] Store page while game is extracting

BrettCleary:
# Summary

Runs the `ExtractZipService` in a worker thread b wrapping `ExtractZipService` with `ExtractZipServiceWrapper`

Key benefits:
- `process.noAsar` can be modified without affecting the main thread
- performance improvements since extraction maxes out the main thread core 

closes https://github.com/HyperPlay-Gaming/px-pm/issues/112


flavioislima (approved):
Code looks good, I believe it is important to test this on all platforms and with the artifact just to be sure it won't break.

PR #1255, opened 2025-03-10
[Chore] update MetaMask logo in the navbar and wallet connection modal

biliesilva:
## Changes
- Added MetaMask icon to the TopNavBar component
- Integrated MetaMask icon from @hyperplay/ui component library
- Icon is displayed when MetaMask browser sidebar links are enabled
- Added styling for the MetaMask icon through the CSS module

## Technical Details
- Used Images.MetaMask component with appropriate styling
- Icon visibility is controlled by showMetaMaskBrowserSidebarLinks context
- Maintains existing badge notification functionality

### Changes
image

image

---


src/frontend/components/UI/TopNavBar/index.module.scss
```diff
@@ -73,3 +73,8 @@
     display: none;
   }
 }
+
+.metaMaskIcon {
+  width: var(--space-md);
+  height: var(--space-md);
```
BrettCleary, line 79: This will go from 16 to 14px at the 1300px breakpoint. Is that what you intended?

I think it's fine to just hardcode here or use `--space-md-fixed` if you really want to use a value from the spacing system.

It also might be best to just set the width here. I think it is being slightly distorted by setting both since the svg aspect ratio isn't 1:1

PR #1259, opened 2025-03-13
[FIx] Edge case where downloaded size is bigger than download size

flavioislima:
This changes how we get the Download Size depending on the status, right now it uses download status for both install and extracting. 

It also adds a safeguard to the Download Toast so it won't show a Downloaded size bigger than Total Download Size. Might not be needed but it is 'just in case' since our numbers might not be precise sometimes.

### Current issue
image
image

### Fixed
https://github.com/user-attachments/assets/dfc178a2-93d0-463a-b349-991159cdb7f9

---

### DIFF
--- a/src/frontend/components/UI/DownloadToastManager/index.tsx
+++ b/src/frontend/components/UI/DownloadToastManager/index.tsx
@@ -14,6 +14,7 @@ import DMQueueState from 'frontend/state/DMQueueState'
 import { isNotNative } from 'frontend/helpers/library'
 import { hasProgress } from 'frontend/hooks/hasProgress'
 import { useGameInfo } from 'frontend/hooks/useGameInfo'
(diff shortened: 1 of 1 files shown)


biliesilva (approved):
LGTM!

BrettCleary:
@flavioislima do you know why the amount download (2.1GB) is greater than the max to download (2.05GB) here? 
image

flavioislima:
> @flavioislima do you know why the amount download (2.1GB) is greater than the max to download (2.05GB) here? 
> image
> 

The second part, smaller one, was using download size, while the downloaded, first part, was using install size.

You can see on the game page that the calculation is correct but here was wrong.

nyghtstalker (approved):
Confirmed that the download size does not exceed the expected amount for both download and extraction.

image
image
image

PR #1261, opened 2025-03-20
[FIX] Reload app due to extension only on app update 

flavioislima:
This is a new approach from #1249.

### AI Summary
This pull request includes several changes to the `src/backend/main.ts` and `src/common/types/electron_store.ts` files to handle app version updates and ensure proper functionality when the app version changes. The changes primarily focus on checking the app version and reloading the app if the version has changed.

### Handling app version updates:

* [`src/backend/main.ts`](diffhunk://#diff-859af3c776da914ebff60fa13adfc96e79d99ea563fe1e4e790de14656442accR473-R483): Added logic to compare the current app version with the stored version and log the change. If the app version has changed, the new version is stored.
* [`src/backend/main.ts`](diffhunk://#diff-859af3c776da914ebff60fa13adfc96e79d99ea563fe1e4e790de14656442accR720-R735): Added a check in the `ipcMain.once('frontendReady', async () => {` block to reload the app if the app version has changed.

### Type definitions:

* [`src/common/types/electron_store.ts`](diffhunk://#diff-1e663c824f2fa561f595dca3a54d1be51a0be527bf080539216372c5697b3d5bR24): Added `appVersion` to the `StoreStructure` interface to store the app version.

---

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -470,6 +470,17 @@ if (!gotTheLock) {
 
     createInjectedProviderWindow()
 
+    const currentStoredVersion = configStore.get('appVersion', '')
+    const currentVersion = app.getVersion()
+
+    if (currentStoredVersion !== currentVersion) {
+      logInfo(
+        `App version changed from ${currentStoredVersion} to ${app.getVersion()}`,
+        LogPrefix.Backend
+      )
+      configStore.set('appVersion', currentVersion)
+    }
+
     const providerPreloadPath = path.join(
       __dirname,
       '../preload/providerPreload.js'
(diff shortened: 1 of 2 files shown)


biliesilva (approved):

src/backend/main.ts
```diff
@@ -706,10 +717,31 @@ ipcMain.once('loadingScreenReady', () => {
...
+
   await initExtension(hpApi)
-  // wait for mm SW to initialize
-  await wait(5000)
-  ipcMain.emit('reloadApp')
+
+  // Only reload the app if a wallet is not connected
+  // or the app version has changed
+  if (
+    !walletStateIsConnected ||
```
BrettCleary, line 732: which edge case does this address?

flavioislima: Joe found out that it is not possible to import browser profiles if we do not reload. But looking at the logic not, this issue was maybe caused by the order on which I was adding the initExtension before. I will ask him to test again and will remove this check for now.

nyghtstalker (approved):
- I was able to reset the extension and select my chrome MM extension profile now!
- After changing the appVersion I was still logged into my MM extension profile
- I reset again and was still able to log into my MM extension profile

src/backend/main.ts
flavioislima: OK it seems we did not need this.

src/backend/main.ts
```diff
@@ -2026,6 +2050,44 @@ ipcMain.handle(
...
+    if (accounts && accounts.length > 0) {
+      configStore.set('walletState.address', accounts[0])
+      configStore.set('walletState.provider', provider)
+      configStore.set('walletState.isConnected', true)
+    } else {
+      configStore.set('walletState.isConnected', false)
+    }
+  }
+)
+
```
BrettCleary, line 2090: wait why did we add these?

src/common/types/electron_store.ts
```diff
@@ -43,6 +44,11 @@ export interface StoreStructure {
     'window-props': Electron.Rectangle
     settings: AppSettings
     skipVcRuntime: boolean
+    walletState: {
+      isConnected: boolean
+      address: string
+      provider: PROVIDERS
+    }
```
BrettCleary, line 51: do we need this?

BrettCleary (commented):
I think there are still some leftover changes from the previous approach

src/backend/main.ts
flavioislima: Yeah, I didnt want to remove those because I believe this is useful information for us to have access on backend of frontend. It basically just stores the current account and the state, so we can use to have analytical inputs.
But I can remove it if you think we won't use it.

src/common/types/electron_store.ts
flavioislima: For the reasons above

src/backend/main.ts
BrettCleary: Yeah let's remove it since we aren't using it right now

flavioislima: done

src/common/types/electron_store.ts
flavioislima: code removed

PR #1272, opened 2025-04-23
[Chore] Remove Error updater dialog

flavioislima:
Show error dialog only for app updates errors and not for network changed, and similar temporary errors.

### How to test

- Hard to test since it will only affect specific errors codes during update check.

---


biliesilva (approved):

BrettCleary:
> it is redundant now that we have the alerts

which alerts are you referring to here @flavioislima ?

BrettCleary:
> > > it is redundant now that we have the alerts
> > 
> > 
> > which alerts are you referring to here @flavioislima ?
> 
> Mixpanel alerts. These error dialogs on the client are not really helpfull.

Wouldn't this create silent failures client-side when a user is auto-updating though?

If a user doesn't know auto-update failed, then they won't think to go and manually update

flavioislima:
> > > > it is redundant now that we have the alerts
> > > 
> > > 
> > > which alerts are you referring to here @flavioislima ?
> > 
> > 
> > Mixpanel alerts. These error dialogs on the client are not really helpfull.
> 
> Wouldn't this create silent failures client-side when a user is auto-updating though?
> 
> If a user doesn't know auto-update failed, then they won't think to go and manually update

I changed to code to still show the error message to some error codes. 

But looking at how other Electron apps behaves, check VScode, Discord and others. They won't show error messages for those. In 99% of the cases the app will eventually update unless you cannot download the update at all. So this type of dialog we have is pretty useless and most people won't understand it.

PR #1289, opened 2025-05-20
[Feat] Updates to quest reward claiming

BrettCleary:
# Summary

pulls in https://github.com/HyperPlay-Gaming/quests-ui/pull/95

# Testing

https://app.qase.io/case/HC-587


src/backend/ipcHandlers/achievements.ts
```diff
@@ -254,6 +255,14 @@ async function getQuestRewardSignature(
     url: DEV_PORTAL_URL
   })
 
+  let hwAttestation: HardwareAttestation | undefined = undefined
+  try {
+    const extProvider = await import('@hyperplay/extension-provider')
+    hwAttestation = await extProvider.getHwSig()
+  } catch (err) {
+    logError(`Error getting hardware signature ${err}`)
```
eliobricenov, line 263: What do you think about adding an event for this?

eliobricenov (approved):

src/backend/ipcHandlers/achievements.ts
BrettCleary: should already get captured in the on claim top level error handler in quests-ui

BrettCleary:
ran the test again and it is working. merging now

PR #1300, opened 2025-05-30
[Fix] Add action rejected check to set active wallet

BrettCleary:
# Summary

We were returning an object on set active wallet sign rejection and sending {} to the api for signature causing InvalidSignature during the set active wallet flow

closes https://github.com/HyperPlay-Gaming/hyperplay-pm/issues/578


jiyuu-jin (approved):
LGTM

PR #1319, opened 2025-06-24
[Chore] Update MetaMask to v12.20.1

BrettCleary:
# Summary

closes https://github.com/HyperPlay-Gaming/hyperplay-pm/issues/596


biliesilva (approved):
Tested on Mac and no spinning problem. it opens MetaMask extension right after the Store, and I could login smothly.

PR #1320, opened 2025-06-25
[Feat] add support component and zendesk script

BrettCleary:
# Summary

closes https://github.com/HyperPlay-Gaming/hyperplay-pm/issues/618

# Testing

https://app.qase.io/case/HC-597

other tests were not added to qase as they would be re-testing cases already handled by zendesk


eliobricenov (approved):
worked great for me!

PR #1322, opened 2025-07-08
Feat/implement new connect wallet modal text/graphs

biliesilva:

Closes: https://github.com/orgs/HyperPlay-Gaming/projects/16/views/8?pane=issue&itemId=115770505&issue=HyperPlay-Gaming%7Chyperplay-pm%7C633

Design: 
image

After: 
image

---

### DIFF
--- a/src/frontend/screens/Onboarding/walletSelection/screens/info/index.tsx
+++ b/src/frontend/screens/Onboarding/walletSelection/screens/info/index.tsx
@@ -5,6 +5,12 @@ import { ConnectWallet, WarningIcon } from 'frontend/assets/hyperplay'
 
 import { t } from 'i18next'
 
+interface WalletInfoScreenProps {
+  skipClicked: () => void
+  createWalletClicked: () => void
+  mmInitialized: boolean
+}
+
 interface InfoTextProps {
   icon: React.ReactNode
   title: string
@@ -25,7 +31,8 @@ const InfoText = function ({ icon, title, details }: InfoTextProps) {
   )
 }
 
-const WalletInfoScreen = () => {
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const WalletInfoScreen = (props: WalletInfoScreenProps) => {
   return (
     <>
       <div className={WalletInfoStyles.walletInfoScreen}>


pimentel08:
Let's remove the "Don't have a wallet" section

src/frontend/screens/Onboarding/walletSelection/screens/info/index.tsx
```diff
@@ -28,66 +31,84 @@ const InfoText = function ({ icon, title, details }: InfoTextProps) {
...
-            'hyperplay.onboarding.walletSelection.screens.info.skip',
-            `Skip for now`
+            'hyperplay.onboarding.walletSelection.screens.info.digitalAssets.smallTitle',
+            `Connect your wallet to activate powerful features:`
           )}
-        </Button>
-        {!props.mmInitialized ? (
-          <Button onClick={props.createWalletClicked}>
+        </div>
+        <div className={`body-sm ${WalletInfoStyles.description}`}>
```
BrettCleary, line 54: looks like `body-sm` is 14px in figma and 16px in hp/ui. we should fix that. I think it is one of the reasons the height of this section is larger than the figma design

src/frontend/screens/Onboarding/walletSelection/screens/info/index.tsx
```diff
@@ -28,66 +31,84 @@ const InfoText = function ({ icon, title, details }: InfoTextProps) {
...
+          <li>
+            {' '}
```
BrettCleary, line 56:
```suggestion
```

src/frontend/screens/Onboarding/walletSelection/screens/info/index.tsx
```diff
@@ -28,66 +31,84 @@ const InfoText = function ({ icon, title, details }: InfoTextProps) {
...
+            target="_blank"
+            className={WalletInfoStyles.createWalletLink}
+            onClick={props.createWalletClicked}
+          >
+            {t(
+              'hyperplay.onboarding.walletSelection.screens.info.createWalletLink',
+              `Get MetaMask`
+            )}
+          </a>
+        </div>
```
BrettCleary, line 100:
```suggestion
```
@pimentel08 suggested to remove this

BrettCleary (commented):
we should look into why there is a height difference with this section vs figma. also there is the difference in background color

this pr
image

figma 
image

biliesilva:
> Also, it seems like there is a font size discrepancy with the Alert text. Could we have that updated, pls?
> 
> Screenshot 2025-07-08 at 3 49 01 PM

@pimentel08, in the design for this alert text, we have Barlow, which is a font that we don't use anymore. In code, I'm using eyebrow-sm from our typography from our ui-lib. Also, could you confirm that we will have the change also in the background color of this modal, like @BrettCleary commented above? Seems like it will differ from the other background colors of our other modals, and it is black 00000 in the design, not following our DS colors. Seems like the design is not all updated with our DS.

eliobricenov (approved):

PR #1324, opened 2025-07-09
[Tech] Update hp ui to new exports

BrettCleary:
# Summary

Brings in latest changes from hp ui so we don't lose trunk based development

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hyperplay",
-  "version": "0.30.0",
+  "version": "0.30.1",
   "private": true,
   "main": "build/main/main.js",
   "homepage": "./",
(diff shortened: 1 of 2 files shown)


src/frontend/screens/Achievements/index.tsx
```diff
@@ -1,8 +1,11 @@
 import React from 'react'
 import walletState from 'frontend/state/WalletState'
-import { AchievementCard, AchievementSummaryTable } from '@hyperplay/ui'
+import {
+  AchievementCardProps,
+  AchievementCard,
+  AchievementSummaryTable
+} from '@hyperplay/ui'
 import { NavLink } from 'react-router-dom'
-import { StatusIconState } from '@hyperplay/ui/dist/components/AchievementCard/components/StatusIcon'
```
BrettCleary, line 5: we should be more strict on PR reviews about imports like this

BrettCleary: or add a ts/eslint rule against it if possible

PR #1335, opened 2025-09-24
Fix/traverse via controller

iabubakardev:
- We can now easily traverse between the topbar and sidebar.
- From the sidebar, focus can move into the screens of Library, Download Manager, and Settings, and all elements inside are traversable.
- Returning to the sidebar is simple using LeftPad.
- Adjusted deltaX and deltaY to maximize joystick reach when moving between elements.
- Fixed the MetaMask modal issue: pressing Enter now keeps the modal active. To close it, use Esc (I’m not sure what the equivalent is on a controller).
- Fixed the border color around the Discord and X buttons in the sidebar.


BrettCleary:
@iabubakardev  When I have the metamask icon in the nav bar focused, I click right on the d pad and it goes left to the search bar

image

and then after clicking right on d pad
image

iabubakardev:
@BrettCleary I'll have another look.

